PR #4, opened 2019-08-14
Change paybutton to Bitcoin Cash green

alt-win:
This simply changes the paybutton button and modal colors to Bitcoin Cash green (previously orange).
Everything else is left untouched.


Klakurka:
Green is now the default. Orange can be set with `theme: "orange"`.

Issue #27, opened 2020-07-20
Add option to show 'goal' progress bar


Klakurka:
Add option to show 'goal' progress bar.

Klakurka:
Original concept for this: 
![image](https://user-images.githubusercontent.com/4713204/95391557-1a165380-08ac-11eb-8839-cde82976ab56.png)

Klakurka:
Add a goal field that adds a progress bar below the button and in the dialog and prevents the dialog from entering a success state (instead the progress bar updates).

Klakurka:
Example styling that would work great for the widget'

https://www.reddit.com/r/btc/comments/kt0snz/you_asked_and_we_listened_you_can_now_leave_your/

PR #77, opened 2021-02-26
27 Add option to show 'goal' progress bar

paOol:
https://github.com/PayButton/paybutton/issues/27


Klakurka:
- Let's change how it looks a bit:
  - 
image
  - `margin-bottom: 0.6em` on the top text.
  - Remove 'goal' text on the bottom.
  - Use `margin-bottom: 20px` instead of 25px on the progress bar. 
  - Change the top text to `Y.00 of X.00 <strong>BCH</strong>` (2 decimals is probably good, bolded BCH). This matches up with what Flipstarter does.
 - Can we put the loading spinner/animation in place of the progress bar section while it's loading in? We want to stop the dialog/widget from jumping down when the progress bar appears.

Klakurka (changes requested):
- Let's change how it looks a bit:
   - 
image
   - `margin-bottom: 0.6em` on the top text.
   - Remove 'goal' text on the bottom.
   - Use `margin-bottom: 20px` instead of 25px on the progress bar. 
   - Change the top text to `Y.00 / X.00 <strong>BCH</strong>` (2 decimals is probably good, bolded BCH). This matches up with what Flipstarter does.
 - Can we put the loading spinner/animation in place of the progress bar section while it's loading in? We want to stop the dialog/widget from jumping down when the progress bar appears.
 - The progress bar should update when a deposit is detected. It currently looks at the confirmed balance but it should include unconfirmed as well.

Klakurka:
- When the amount value is 0, it's showing `Send 0 BCH` at the top instead of `Send any amount of BCH` like before.
- X / Y BCH value still doesn't update on payment, only after 1 confirmation.

Issue #33, opened 2020-10-12
disableEnforceFocus required to avoid Material UI from throwing error when opening with JS 


Klakurka:
Possible solution is to expose `disableEnforceFocus` so the caller can pass in `disableEnforceFocus: true`.

PR #78, opened 2021-02-26
#33: Update docs and switch flag to true by default

mmbutton:


react/src/components/PayButton/PayButton.tsx
```diff
@@ -105,6 +107,11 @@ PayButton.defaultProps = {
   randomSatoshis: true,
   successText: 'Thank you!',
   disabled: false,
+<<<<<<< HEAD
```
Klakurka, line 110: This looks like it got messed up when reversing it arounnd?

mmbutton: Yup, Windows is undoing my changes in VSCode for some reason.

paybutton/rollup.dev.js
```diff
@@ -48,8 +48,13 @@ function template ( { attributes, title, bundle } ) {
           <div class="paybutton" to="bitcoincash:qrmm7edwuj4jf7tnvygjyztyy0a0qxvl7q9ayphulp" amount="50000" currency="SAT" detect-payment="true"</div>
           <br />
           <div class="paybutton-widget" to="bitcoincash:qrmm7edwuj4jf7tnvygjyztyy0a0qxvl7q9ayphulp" amount="50000" currency="SAT" detect-payment="true"</div>
+          <br />
+          <div id="js-test" />
         </div>
       </body>
+      <script>
+        PayButton.renderWidget(document.getElementById('js-test'), { to: 'bitcoincash:qrmm7edwuj4jf7tnvygjyztyy0a0qxvl7q9ayphulp', currency: 'SAT', detect-payment: 'true' });
```
Klakurka, line 56: Wasn't in scope to add this but it doesn't seem to be working either way.

docs/README.md
```diff
@@ -522,6 +522,35 @@ randomSatoshis = true
...
+
+?> This parameter is optional. Default value is true. Possible values are true or false.
+
+**Example:**
+<!-- tabs:start -->
+
+#### ** HTML **
+
+```html
+disableEnforceFocus="false"
```
Klakurka, line 537: For HTML, we have to use lower case (`disable-enforce-focus` or something). I'm not sure how this works exactly as in it might automatically do this on build based on the camelCase... really not sure.

Klakurka (commented):
Did you find a good way to test this one?

Klakurka:
Also let's go ahead and add the readme changes to both of the chinese versions as well (just copied, in English) and we can get those translated later. It'll be easier to see what's "missing" if it's not missing from the docs, just not yet translated.

paybutton/rollup.dev.js
mmbutton: I may as well remove it as I'm going to be fixing the docs anyways.

docs/README.md
mmbutton: I think it's a copy-paste error from me. I'll fix it up

mmbutton:
I looked into storybook with HTML and JS but it would take a good chunk of effort to get it to play nicely with what we currently have. Might be worth looking into in a different ticket but I'm not sure how often we need to test JS and HTML.

I tested it via prop manipulation on storybook & the code in rollup.dev.js (Which didn't seem to change the test page. I ended up testing via the console instead). I didn't test it on a Material UI page.

Klakurka:
Tested the actual functionality and it works.

The docs need some work though:

1. We use the HTML parameter name formatting by default so can you change those over to `disable-enforce-focus`?
2. `_sidebar.md` needs to be updated (both in english and in `zh-cn` and `zh-tw`. Can use `disable-enforce-focus` for all of it.
3. Remove the space before the `**` at the end of the parameter description so that it actually bolds it and changes the font colour to black.
4. Move the `// Note: XXX` to the end of the parameter description. Something like: `The ‘disableEnforceFocus’ parameter is passed to the Dialog material UI component. Setting it to false can help with accessibility with technology such as screen readers but may throw errors on sites running Material UI.`

Klakurka:
There's also a conflict that needs to be resolved apparently.

docs/README.md
Klakurka: This is still an issue CTRL+F on `disableEnforceFocus` and change them to `disable-enforce-focus`. Clicking the sidebar link should take you to this section in the docs.

Klakurka: Will need to update all 3 versions of the README.

mmbutton: Can you recheck the latest version of the README's? The HTML parameter is correct in all of them for me.

Klakurka: Maybe you didn't push? It's still not right here on GitHub or in the version I pulled.

https://github.com/PayButton/paybutton/pull/78/files

mmbutton: When I click the link it shows the HTML version being correct. I'll double check with a clean pull

mmbutton: Ah, are you talking about ## disableEnforceFocus on line 554? The original comment was on the html parameter which has been corrected

Issue #59, opened 2021-01-07
Improve website dark mode


Klakurka:
- Improve readability of text (lighten font colour).
- Have dark mode apply to all of the markdown elements. Currently there's a lot of white.

PR #87, opened 2021-03-02
59 Improve website dark mode

paOol:
https://github.com/PayButton/paybutton/issues/59

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -161,7 +161,7 @@ function App() {
 
 You can also create an always-visible PayButton Widget that doesn't require clicking a button to open:
 
-<div class="paybutton-widget" to="bitcoincash:qrmm7edwuj4jf7tnvygjyztyy0a0qxvl7q9ayphulp"></div>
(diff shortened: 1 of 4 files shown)


paOol:
the docs are generated by docsify rather than plain html/css/js, 
there are various themes and most do not include a light/dark mode.
the paybutton is using the  `docsify-dark-mode` npm package
there was only 3 parameters to pass in color options so I had to "fork" the javascript to enter additional options.
The same had to be done with the CSS.

Klakurka (changes requested):
Much better!

- Can we boost the font colours on the titles, nav, and script tags? 
image
- Could we also darken the background on the "Requires React v16.3 or later." section?
- Is there anything we can do about the widget width? Its width makes it stand out quite a bit in dark mode.
- When dark mode is on, it's making the title text "The easiest way to accept Bitcoin Cash online." harder to read... what do you think about darkening the background on the landing page section when using dark mode? If we did that, the "Language" text in the upper right will need to be lightened and the PayButton logo border will need to be improved as the edges are pretty sharp on a dark background -- although we can bring in somebody else to improve the logo.

PR #92, opened 2021-03-03
88 Ensure 'disabled' property works by itself with a valid address entered

paOol:
https://github.com/PayButton/paybutton/issues/88


Klakurka (changes requested):
What do you think about hiding the `Click to copy` div as well?

image

paOol:
updated to remove the text.

Issue #91, opened 2021-03-02
Stop website styling from leaking into PayButton widget


Klakurka:
As an example, the 'Powered by PayButton.org' font is different on https://paybutton.org than it is in the dialog triggered by the button.

Klakurka:
Looks like this is still an issue: 
![image](https://user-images.githubusercontent.com/4713204/109887574-d6473080-7c36-11eb-8298-78867d5a6b0f.png)

PR #93, opened 2021-03-03
91 Stop website styling from leaking into PayButton widget

paOol:
https://github.com/PayButton/paybutton/issues/91

this depends on the PR from 
https://github.com/PayButton/paybutton/issues/88
to work properly.

the docs also must point to the newly built paybutton.js on unpkg before seeing changes.

### DIFF
--- a/docs/darkmode.js
+++ b/docs/darkmode.js
@@ -48,8 +48,8 @@ const plugin = (hook, vm) => {
     },
     light: {
       background: 'white',
-      toggleBtnBg: 'var(--theme-color)',
-      textColor: 'var(--theme-color)',
+      toggleBtnBg: '#34495e',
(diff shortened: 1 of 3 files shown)


Klakurka (changes requested):
Website styling is fixed, looks good.

We're still getting some style leaking into the widgets. Compare how the "Powered by PayButton.org" text looks on the widget vs in the dialog.

The intention with this task is to (ideally) have PayButton be resilient to outside styles, if possible.

So to fix this, it's probably better to have PayButton itself "fend off" the docs website styles rather than updating the docs styling to not touch the widget.

paOol:
the style leaking requires changes from this PR 
https://github.com/PayButton/paybutton/pull/92

Issue #96, opened 2021-03-04
Rollup build failing


Klakurka:

Klakurka:
Still an issue looks like.

Klakurka:
Switching to Webpack is on the table as a possible option.

Ideally, we'd want to be able to run 1 command hand have it build both the React and non-React versions.

Klakurka:
"Fixed" with this: https://github.com/PayButton/paybutton/pull/100.

PR #97, opened 2021-03-04
replace master with working build

paOol:
https://github.com/PayButton/paybutton/issues/96


paOol:
closing because it merged previous bad commits into this

Issue #111, opened 2021-03-09
Fiat price sometimes changes shortly after the dialog is up


Klakurka:
Context: https://github.com/PayButton/paybutton/pull/101#issuecomment-793529595

Issue #105, opened 2021-03-06
Ignore values from bad API calls so we don't get NaNs showing up


Klakurka:
![image](https://user-images.githubusercontent.com/4713204/110220762-a38e7980-7e7c-11eb-8a32-336f49478359.png)

Klakurka:
![image](https://user-images.githubusercontent.com/4713204/110221798-0d118680-7e83-11eb-8119-deab09f63df0.png)

paOol:
defaults to 'Send any amount' message when receiving a NaN value

Klakurka:
Just want to double check what happens when bad data comes in when `goalAmount` is set.

PR #114, opened 2021-03-10
111 Fiat price sometimes changes shortly after the dialog is up

paOol:
https://github.com/PayButton/paybutton/issues/111


paOol:
possibly addresses https://github.com/PayButton/paybutton/issues/105 as well.

Issue #75, opened 2021-02-25
Show better error message in the console when unable to create the button


Klakurka:
Right now it just throws some obscure error about a variable, making it look like it's the `paybutton.js` that's broken rather than something simple as being unable to find the creation div ID (perhaps because it wasn't ever added by the web developer).

paOol:
paybutton/paybutton/src/index.tsx

line 105. 

leaving notes for myself

Issue #18, opened 2020-02-11
Add pay-to address error checking / validation


Klakurka:
Add pay-to address error checking, making it obvious that there's an issue with the address, cashaccount, etc if it's entered incorrectly.

Klakurka:
This is supported in bitcoin-ts:

https://twitter.com/bitjson/status/1226944424338481155

Klakurka:
Still an issue.

Klakurka:
Add address validation that will disable (#23) the button, and hide/grey-out the widget if a valid address isn't being used.

Klakurka:
Ideally this should use the same code path as #23 would.

Klakurka:
In fact, it might even make sense to do #23 at the same time (your call, whatever makes sense).

PR #115, opened 2021-03-11
75 Show better error message in the console when unable to create the button

paOol:
https://github.com/PayButton/paybutton/issues/75

renders the component 
displays errors
disables button

### DIFF
--- a/paybutton/src/index.tsx
+++ b/paybutton/src/index.tsx
@@ -4,7 +4,7 @@ import { h } from 'preact';
 import { render } from 'preact/compat';
 
 declare global {
-  interface Window { 
+  interface Window {
     WebKitMutationObserver: any;
   }
 }
@@ -23,7 +23,7 @@ function init ( ) {
 
   const MutationObserver = window.MutationObserver ?? window.WebKitMutationObserver;
   const observer = new MutationObserver( render );
-  observer.observe( document, { 
+  observer.observe( document, {
     subtree: true,
     childList: true,
     attributes: true,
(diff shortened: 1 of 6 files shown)


paybutton/src/index.tsx
```diff
@@ -104,7 +104,7 @@ function findAndRender <T>( className: string, Component: React.ComponentType<an
 
       if ( ! requiredProps.every( name => name in attributes ) ) {
         console.error( 'PayButton: missing required attribute: ' + JSON.stringify( requiredProps.filter( name => ! ( name in attributes ) ) ) );
-        return;
+        // return;
```
Klakurka, line 107: Meant to be commented out still?

Klakurka (changes requested):
What's the main functional difference with this one?

The goal was to make it clearer to the user what is wrong when they've made mistakes in the set up.

For example,

1. Can't find element ID on page (either missing or misspelt).
2. Garbage entered for any of the parameters.

Ideally, we'd say "Hey, X parameter was set incorrectly. Please check it." in the console.

They currently see an unhelpful message. Eg. here is what happens when it can't find the element ID of the div to create the button in:

image

paOol:
The original flow with a missing "to" param was to not render anything and just have a console warning. 

Now the button renders (disabled) and the erorr msg is shown in the component as well as the console. 
Other errors can also be shown this way to make them more visible.


`can't find the element ID of the div to create the button in:`

I haven't tested this yet but can include it into this PR.

paybutton/src/index.tsx
paOol: yes,  for someone in the future to be able to see that and easily revert if needed. versus having to dig through git history.

Klakurka:
Yeah, include that.

It's supposed to also be able to handle garbage inputs. Anytime there's something that would throw an error (because of a bad input), it would let them know which parameter is messed up.

paOol:
testing should be done by building and creating via HTML or JS.

storybook is clunky to force bad values.

Klakurka:
image

I get this when I view it on localhost:10001 and remove the `to` field.

Klakurka:
A few things seem to be broken... not sure what the deal is:

 - 'Click to copy' isn't working for me.
 - Address validation isn't working for me: https://github.com/PayButton/paybutton/issues/18. If I enter `bitcoincash:asqwefsadfsad`, it still tries to use it. 
image


Tested on `http://localhost:10001/`.

PR #116, opened 2021-03-17
105 Ignore values from bad API calls so we don't get NaNs showing up

paOol:
https://github.com/PayButton/paybutton/issues/105


Klakurka:
Not sure this is what we want to do... If there are 3-4 bad api calls in a row, due to a poor network connection or something, we probably want to simply ignore it if we have a good value already.

Disabling the element is what we want to do if we haven't gotten any good data yet. Is there a way to cover both cases?

Issue #86, opened 2021-03-01
Parameter for allowing individuals to edit the amount


Klakurka:
A small edit/pencil button would be visible somewhere.

If amount is null/0: 
![image](https://user-images.githubusercontent.com/4713204/109568314-98fa6b80-7a9b-11eb-8d38-941d3347c1ee.png)

If amount is >0: 
![image](https://user-images.githubusercontent.com/4713204/109568276-86803200-7a9b-11eb-87d9-f909cd8cfd37.png)

Klakurka:
I actually don't mind this.

Feedback:

 - We should add an error message for negative numbers. Text should be `Amount should be positive`.
 - Pencil can be maybe reduced in size from 30 -> 20.
 - Add input character limit to input of maybe 12.
 - If input is empty, set the value to 0, which would also trigger the new `Amount should be positive` error.
 - Error message location could be moved from the bottom to the top bar section rather than it displaying this: 
![image](https://user-images.githubusercontent.com/4713204/112397477-653beb80-8cbf-11eb-8322-a65067bdb562.png)

paOol:
one note, 
I think the default behavior (0 amount) should show `Send any amount of BCH` and not an error state.

Klakurka:
Yep, you're right.

PR #117, opened 2021-03-25
86 Parameter for allowing individuals to edit the amount

paOol:
https://github.com/PayButton/paybutton/issues/86


Klakurka:
- Add input character limit to input of maybe 12.
- When randomSatoshis = true, setting the amount to 0 should also have the randomSatoshis return 0. Right now if you set the value to 0 with randomSatoshis = true, it'll show "send 0.00007384 BCH" (as an example).

Klakurka:
- 
image This happened when I updated the input, then deleting the amount in the input (which sets it to 0). The top section didn't update back to 'Send any amount'.

Klakurka:
- When the amount input value is 0, it should say 'Enter any amount' instead of 'Send 0.00000000 BCH'.
- Let's change the property name from `userCanEdit` to `editable` to be more in line with the other names.

Klakurka (changes requested):
- When the amount input value is 0, it should say 'Enter any amount' instead of 'Send 0.00000000 BCH'.
- Let's change the property name from `userCanEdit` to `editable` to be more in line with the other names.

react/src/components/Widget/Widget.tsx
```diff
@@ -289,7 +289,9 @@ export const Widget: React.FC<WidgetProps> = props => {
       url = prefixedAddress + (query.length ? `?${query.join('&')}` : '');
       setUrl(url);
     } else {
-      if (!isFiat && currencyObj) {
+      const notZeroValue: boolean =
+        currencyObj?.satoshis !== undefined && currencyObj.satoshis > 0;
```
Klakurka, line 293: `currencyObj?.satoshis` I'm not familiar with what the '?' does here. Able to explain?

PR #118, opened 2021-03-28
10 Add (optional) server/api with address checking and notifications

paOol:
replaced rest, so it's important to be thorough to Q/A for issues.

### DIFF
--- a/react/src/hooks/useAddressDetails.ts
+++ b/react/src/hooks/useAddressDetails.ts
@@ -3,7 +3,7 @@ import { useEffect, useState } from 'react';
 import { AddressDetails, getAddressDetails } from '../util/api-client';
 import { validateCashAddress } from '../util/address';
 
-const POLL_DELAY = 100;
+const POLL_DELAY = 1000;
 
 export const useAddressDetails = (
   address: string,
--- a/react/src/util/api-client.ts
+++ b/react/src/util/api-client.ts
@@ -11,26 +11,25 @@ import axios from 'axios';
 
 export const getAddressDetails = async (
   address: string,
+  rootUrl = `https://api.paybutton.org`,
 ): Promise<AddressDetails> => {
-  const res = await fetch(
-    `https://api.paybutton.org/address/transactions/${address}`,
-  );
+  const res = await fetch(`${rootUrl}/address/transactions/${address}`);
   return await res.json();
 };
 
 export const getSatoshiBalance = async (
   address: string,
-  rootUrl = `https://api.paybutton.org/`,
+  rootUrl = `https://api.paybutton.org`,
 ): Promise<{ satoshis: number }> => {
-  const res = await fetch(`${rootUrl}address/balance/${address}`);
+  const res = await fetch(`${rootUrl}/address/balance/${address}`);
   return await res.json();
 };
 
 export const getUTXOs = async (
   address: string,
-  rootUrl = `https://api.paybutton.org/`,
+  rootUrl = `https://api.paybutton.org`,
 ): Promise<UtxoDetails> => {
-  const res = await fetch(`${rootUrl}address/utxo/${address}`);
+  const res = await fetch(`${rootUrl}/address/utxo/${address}`);
   return await res.json();
 };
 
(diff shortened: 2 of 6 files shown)


react/src/hooks/useAddressDetails.ts
```diff
@@ -3,7 +3,7 @@ import { useEffect, useState } from 'react';
 import { AddressDetails, getAddressDetails } from '../util/api-client';
 import { validateCashAddress } from '../util/address';
 
-const POLL_DELAY = 100;
+const POLL_DELAY = 1000;
```
Klakurka, line 6: What's `POLL_DELAY` for?

react/src/util/api-client.ts
```diff
@@ -1,14 +1,37 @@
...
+    `https://api.paybutton.org/address/transactions/${address}`,
   );
   return await res.json();
 };
 
+export const getSatoshiBalance = async (
+  address: string,
+): Promise<{ satoshis: number }> => {
+  const res = await fetch(
+    `https://api.paybutton.org/address/balance/${address}`,
```
Klakurka, line 25: Can we move `https://api.paybutton.org/` to a parameter where the default value is `https://api.paybutton.org/` ?

That way it's all in 1 place and can be changed easily by people who decide to run their own server.

react/src/util/api-client.ts
```diff
@@ -1,14 +1,37 @@
...
+  );
+  return await res.json();
+};
+
+export const getUTXOs = async (address: string): Promise<UtxoDetails> => {
+  const res = await fetch(`https://api.paybutton.org/address/utxo/${address}`);
```
Klakurka, line 31: Same here.

react/src/util/api-client.ts
```diff
@@ -19,11 +42,20 @@ export const getFiatPrice = async (currency: currency): Promise<PriceData> => {
...
+//   );
+//   return await res.json();
+// };
+
 export const getTransactionDetails = async (
   txid: string,
 ): Promise<TransactionDetails> => {
   const res = await fetch(
-    `https://rest.bitcoin.com/v2/transaction/details/${txid}`,
+    `https://api.paybutton.org/transactions/details/${txid}`,
```
Klakurka, line 58: And here.

Klakurka (changes requested):

react/src/util/api-client.ts
```diff
@@ -20,15 +20,17 @@ export const getAddressDetails = async (
 
 export const getSatoshiBalance = async (
   address: string,
+  rootUrl = `https://api.paybutton.org/`,
```
Klakurka, line 23: What do you think about moving that last `/` into the `fetch()` calls? Do you think that's better or worse?

gnarj: I think that's a much better way to do it. I'll get that changed asap.

react/src/util/api-client.ts
```diff
@@ -1,14 +1,39 @@
...
+//   return await res.json();
+// };
+
 export const getAddressDetails = async (
   address: string,
 ): Promise<AddressDetails> => {
   const res = await fetch(
-    `https://rest.bitcoin.com/v2/address/details/${address}`,
+    `https://api.paybutton.org/address/transactions/${address}`,
   );
```
gnarj, line 17: @Klakurka Should we change this url to be a parameter as well?

Klakurka: Ya anything that uses the api should be I think.

Klakurka:
Code looks good, just need to test it.

Klakurka:
I might be testing incorrectly but I tried running the normal JS version using rollup and it's still pointing to rest.bitcoin.com.

1. cd paybutton && npm i
2. npm start
3. Site will be running on localhost:10001.

Klakurka (changes requested):
Rollup way isn't updating to use api.paybutton.org for me for some reason.

gnarj:
I was able to solve this by running:
cd paybutton/react
npm run build

Klakurka:
I get this at build time:

image

PR #121, opened 2021-04-29
extracted dialog component and created openDialog function

gnarj:
payment dialog can be rendered independent of button by calling openDialog({ to: 'address' })


Klakurka:
Can you add a test/example into both Storybook and rollup?

It'll allow us to more easily test this both now and in the future.

Klakurka (changes requested):
Can you add a test/example into both Storybook and rollup?

It'll allow us to more easily test this both now and in the future.

gnarj:
@Klakurka added those examples/tests to the rollup and storyboard. if you have any requests on the storyboard examples, i am happy to add them.

Klakurka (changes requested):
We decided that we're going to refactor this a bit but functionality looks good.

Issue #81, opened 2021-02-28
Refactor Widget/WidgetContainer


Klakurka:
Recommendation by @paOol as this is likely causing us some issues.

paOol:
![image](https://user-images.githubusercontent.com/5941389/109434193-f376c780-79c8-11eb-9000-4109b0e7bc0d.png)

WidgetContainer contains logic is that inaccessible when you instantiate the <Widget> Component alone. 

This requires workarounds like duplicating logic from Widgetcontainer into Widget.

gnarj:
From @paOol "the goal would be to "merge" widget and widgetcontainer into one, ideally".

PR #147, opened 2022-06-03
Prevent extra price api calls

johnkuney:
Removing what I believe are superfluous api calls in the Widget component. Its parent component already has a similar function and passes the price data to Widget as a prop, so seems unnecessary here

### DIFF
--- a/react/src/components/Widget/Widget.tsx
+++ b/react/src/components/Widget/Widget.tsx
@@ -239,10 +239,6 @@ export const Widget: React.FC<WidgetProps> = props => {
       typeof amount === 'string' && amount.startsWith('-');
     let cleanAmount: any;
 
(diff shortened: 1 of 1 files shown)


Klakurka:
It does 2 API calls for me now, which is an improvement but still shouldn't be doing 2.

The widget seems to be doing 0 API calls now...

I did my testing using the rollup localhost:10001.

Klakurka (changes requested):
It does 2 API calls for me now, which is an improvement but still shouldn't be doing 2.

The widget seems to be doing 0 API calls now...

I did my testing using the rollup localhost:10001.

Klakurka:
I'm thinking this might be related: https://github.com/PayButton/paybutton/issues/81

PR #154, opened 2022-07-19
chore: updating dependencies to fix broken builds

alcipir:
WIP: client building is breaking randomly with rollup. Mostly on Windows 10/11 machines. 

Test Plan:
```
cd react
npm i && npm run build
cd ../paybutton
npm i && npm run build
(optional) npm start
```

Ideally, running those commands on Windows with NodeJS (tested on LTS and later) should not break.


Klakurka:
It compiles for me now but I get this when I actually try to use the `paybutton.js` it generates:

image

Klakurka (changes requested):
It compiles for me now but I get this when I actually try to use the `paybutton.js`, we get en error in the console and it seem to work at all.

Klakurka (approved):

PR #156, opened 2022-07-29
feat: add dynamic api url through env vars

alcipir:
Description: adds dynamic `rootUrl` through environment variables to the react component. This adds the ability to build the client with different URLs easily by changing the .env file's `REACT_APP_API_URL` variable.

Example: we can test the client against paybutton-server running locally by setting the .env file to:
`REACT_APP_API_URL=http://localhost:3000/api` and then rebuilding the client.

### DIFF
--- a/paybutton/package-lock.json
+++ b/paybutton/package-lock.json
@@ -13,7 +13,7 @@
         "@rollup/plugin-commonjs": "13.0.2",
         "@rollup/plugin-html": "0.2.4",
         "@rollup/plugin-image": "2.1.1",
-        "@rollup/plugin-json": "^4.1.0",
+        "@rollup/plugin-json": "4.1.0",
         "@rollup/plugin-node-resolve": "8.4.0",
         "@rollup/plugin-replace": "2.4.2",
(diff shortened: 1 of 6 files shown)


Klakurka (changes requested):
Good idea but doesn't seem to work for me when I set it to `localhost:3000`.

1. It basically ignores the REACT_APP_API_URL value and just uses whatever the base url is in the browser (eg. file:///C:/Users/XXX/paybutton)
2. The URL literally has `%7Bprocess.env.REACT_APP_API_URL%7D/` in the URL.

Klakurka:
Rerquires completion of https://github.com/PayButton/paybutton-server/issues/195 and merge of https://github.com/PayButton/paybutton-server/pull/186 before we can continue.

Klakurka:
I'm getting a `PayButton is not defined` console error in my test file I made here so need to figure out what's going wrong there befire I can review this again.

PayButton.render(document.getElementById('my_button'), config);

Klakurka (approved):
Works on my desktop aside from this issue: https://github.com/PayButton/paybutton/issues/159

PR #169, opened 2022-10-04
DRAFT: Client upgrade attempt

Gattaca1:
not working yet. uploading so scott can see


Klakurka:
Superseded by https://github.com/PayButton/paybutton/pull/171.

If anyone feels like revising this and upgrading a bunch of stuff all at once then we can still do that but it's no longer necessary to build.

PR #171, opened 2022-10-07
Repair Storybook Script

ScottMcDermid:
Using this solution:
https://github.com/storybookjs/storybook/issues/13332#issuecomment-736004964

`cd ./react`
`npm i`
`npm run build`
`npm run storybook`

Should launch storybook without errors.


Klakurka:
I do get this at the start of the build:

image

ScottMcDermid:
> I do get this at the start of the build:
> 
> image

I get the same warnings in `master`. Should be able to ignore them.

Klakurka (approved):
Fixed using Node 16. Still broken in Node 18 but looks like we'll probably switch to using docker for this reason.

PR #174, opened 2022-10-07
Dockerize Storybook

ScottMcDermid:
#### Test docker startup

```sh
docker-compose down
docker volume remove paybutton_react-deps
docker-compose up --build
```

After storybook initializes, you should be able to view it:

http://localhost:6006

#### Review wording of updated storybook documentation

```sh
cd ./docs
npm run serve
```

Ensure any mention of running storybook is updated to use the new initialization instructions.


Klakurka (changes requested):
I get this near the end of the `docker compose up` run:

image

Klakurka (changes requested):
Sweet this worked. Let's update the docs first before merging.

PR #175, opened 2022-11-03
Issue 152: (part 1) - Add xecaddrjs

Gattaca1:
Swapped out bitauth for xecaddr.js. BCH functionality of the button should remain the same (valid addresses should be accepted, invalid should be rejected)


Klakurka (approved):
Looks like it's working to me.

PR #183, opened 2022-11-23
Add eCash theme

johnkuney:
Added a theme option for eCash that will be automatically applied if an ecash address is detected, but it can be overidden if a custom theme option is entered 

test plan: try using an ecash address and messing with the theme as well

### DIFF
--- a/react/src/themes/index.tsx
+++ b/react/src/themes/index.tsx
@@ -72,9 +72,16 @@ const validateColors = (obj: any): Theme => {
   }
 };
 
-export const useTheme = (defaultTheme?: ThemeName | Theme): Theme => {
+export const useTheme = (
+  defaultTheme?: ThemeName | Theme,
+  isXec?: boolean,
+): Theme => {
   const validated =
-    defaultTheme === 'orange' ? defaultTheme : validateColors(defaultTheme);
+    defaultTheme === undefined && isXec === true
+      ? ThemeName.XEC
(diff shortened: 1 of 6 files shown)


react/src/themes/index.tsx
```diff
@@ -3,15 +3,17 @@ import React, { useContext } from 'react';
 import { Theme } from './Theme';
 import orange from './themes/orange';
 import paybutton from './themes/paybutton';
+import xec from './themes/xec';
 
-export const themes: Record<ThemeName, Theme> = { orange, paybutton };
+export const themes: Record<ThemeName, Theme> = { orange, paybutton, xec };
```
Gattaca1, line 8: lets double up on the theme names - orange, green, blue, btc, bch, xec. 
orange and btc would be the same theme, blue and xec, etc.

Gattaca1 (changes requested):

react/src/themes/index.tsx
Gattaca1: JK. new plan: 1 name for each theme, and the names will be 
bcha - orange
bch - green
xec - blue

Gattaca1: nvm that can be its own branch

Gattaca1 (commented):
LGTM

Gattaca1 (approved):

PR #184, opened 2022-11-29
 Issue 152 - adding support for XEC - really minor refactor

Gattaca1:
No functionality change - this is just collating all the places where it lists the currencies, and adding a couple TODOs

### DIFF
--- a/react/src/util/api-client.ts
+++ b/react/src/util/api-client.ts
@@ -54,7 +54,7 @@ export const getFiatPrice = async (currency: currency): Promise<PriceData> => {
 
 export const getTransactionDetails = async (
   txid: string,
-  rootUrl = `https://api.paybutton.org`,
+  rootUrl = `https://api.paybutton.org`, // TODO: don't hardcode this url in
 ): Promise<TransactionDetails> => {
   const res = await fetch(`${rootUrl}/transactions/details/${txid}`);
   return await res.json();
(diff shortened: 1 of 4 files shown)


react/src/util/api-client.ts
```diff
@@ -67,7 +68,9 @@ export default {
 };
 
 export type fiatCurrency = 'USD' | 'CAD' | 'EUR' | 'GBP' | 'AUD';
+export const fiatCurrencies = ['USD', 'CAD', 'EUR', 'GBP', 'AUD'];
 export type cryptoCurrency = 'BCH' | 'SAT' | 'bits' | 'XEC';
+export const cryptoCurrencies = ['BCH', 'SAT', 'bits', 'XEC'];
```
Klakurka, line 73: Is this a temporary duplication?

react/src/util/api-client.ts
```diff
@@ -67,7 +68,9 @@ export default {
...
 export type currency = cryptoCurrency | fiatCurrency;
```
Klakurka, line 75: Is this a temporary duplication?

Klakurka (changes requested):

ScottMcDermid (approved):
Refactor looks good. Currencies and fiat codes definitely deserve their own dedicated constant.

PR #185, opened 2022-11-29
Issue 152 - adding support for XEC - Remove sats and bits as currencies

Gattaca1:
BCH functionality should remain intact - but `BITS` and `SATS` have been removed as currency types

### DIFF
--- a/react/src/components/Widget/Widget.tsx
+++ b/react/src/components/Widget/Widget.tsx
@@ -183,11 +183,7 @@ export const Widget: React.FC<WidgetProps> = props => {
   const query: string[] = [];
   const isMissingWidgetContainer = !totalReceived;
   const addressDetails = useAddressDetails(to, isMissingWidgetContainer);
-  const isFiat: boolean =
-    currency !== 'SAT' &&
-    currency !== 'BCH' &&
-    currency !== 'bits' &&
-    currency !== 'XEC';
+  const isFiat: boolean = currency !== 'BCH' && currency !== 'XEC';
   const hasPrice: boolean = price !== undefined && price > 0;
   let prefixedAddress: string;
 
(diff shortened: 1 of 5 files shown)


Klakurka (changes requested):
I know that this was part of a previous branch that just wasn't caught / tested but we should still fix it:

While up: 
image

While animating away:

image

Klakurka (approved):
Ok I'm going to approve this because the issue is going to get fixed in the next branch.

react/src/components/Widget/Widget.tsx
```diff
@@ -372,29 +364,18 @@ export const Widget: React.FC<WidgetProps> = props => {
   };
 
   useEffect(() => {
-    const inSatoshis = getCurrencyObject(totalSatsReceived, 'SAT');
+    // const inSatoshis = getCurrencyObject(totalSatsReceived, 'SAT');
```
ScottMcDermid, line 367: Do we need to keep this?

react/src/components/Widget/Widget.tsx
```diff
@@ -186,11 +183,7 @@ export const Widget: React.FC<WidgetProps> = props => {
   const query: string[] = [];
   const isMissingWidgetContainer = !totalReceived;
   const addressDetails = useAddressDetails(to, isMissingWidgetContainer);
-  const isFiat: boolean =
-    currency !== 'SAT' &&
-    currency !== 'BCH' &&
-    currency !== 'bits' &&
-    currency !== 'XEC';
+  const isFiat: boolean = currency !== 'BCH' && currency !== 'XEC';
```
ScottMcDermid, line 186: `isFiat` seems like a core function. Might deserve to be in a currency service.
Also I'd expect this to look more like `fiatCurrencies.includes(currency)`. Especially if we end up adding another cryptocurrency.

ScottMcDermid (changes requested):

react/src/components/Widget/Widget.tsx
Gattaca1: I'm not quite sure how to do `fiatCurrencies.includes(currency)` the typescript way yet. The possible currencies themselves are types. Hm. Maybe by importing `isFiat` and type checking it from the utils folder. Alright I have some ideas for this

react/src/components/Widget/Widget.tsx
Gattaca1: I wont tell anyone if you don't

(I fixed it in the next stacked branch, I'll see if i can do it here without fucking up git)

react/src/components/Widget/Widget.tsx
Gattaca1: In the next stacked branch I moved `isFiat` to `util/api-client.ts` which is functionally serving as a currency service

Gattaca1:
> I know that this was part of a previous branch that just wasn't caught / tested but we should still fix it:

Lets bump existing issues / issues that affect both BCH and XEC to their own branch. Trying to keep the scope of this branch purely to adding XEC support

ScottMcDermid (approved):

PR #188, opened 2022-11-30
Issue 152 - adding support for XEC - Actually add xec support

Gattaca1:
This branch actually adds XEC support. Additionally, it (hopefully) finishes the job of converting all calculations to primary units instead of doing calculations in satoshis

Using a XEC address should change the theme, the QR, button text, etc. Selecting a fiat currency and goal should give the proper XEC and BCH amounts.

### DIFF
--- a/react/.env
+++ b/react/.env
@@ -1,3 +1,3 @@
 IMAGE_INLINE_SIZE_LIMIT=1000000
 REACT_APP_API_URL=https://api.paybutton.org
-REACT_APP_API_PRICING_URL=http://localhost:3000
+REACT_APP_API_PRICING_URL=https://api.paybutton.org
--- a/react/.env
+++ b/react/.env
@@ -1,2 +1,3 @@
 IMAGE_INLINE_SIZE_LIMIT=1000000
 REACT_APP_API_URL=https://api.paybutton.org
+REACT_APP_API_PRICING_URL=http://localhost:3000
--- a/react/src/components/Widget/Widget.tsx
+++ b/react/src/components/Widget/Widget.tsx
@@ -24,7 +24,7 @@ import BarChart from '../BarChart/BarChart';
 
 import { getCurrencyObject, currencyObject } from '../../util/satoshis';
 import { useAddressDetails } from '../../hooks/useAddressDetails';
-import { currency, getAddressBalance } from '../../util/api-client';
+import { currency, getAddressBalance, isFiat } from '../../util/api-client';
 import { randomizeSatoshis } from '../../util/randomizeSats';
 import PencilIcon from '../../assets/edit-pencil';
 
@@ -185,7 +185,6 @@ export const Widget: React.FC<WidgetProps> = props => {
   const query: string[] = [];
   const isMissingWidgetContainer = !totalReceived;
   const addressDetails = useAddressDetails(to, isMissingWidgetContainer);
-  const isFiat: boolean = currency !== 'BCH' && currency !== 'XEC';
   const hasPrice: boolean = price !== undefined && price > 0;
   let prefixedAddress: string;
 
@@ -300,7 +299,7 @@ export const Widget: React.FC<WidgetProps> = props => {
     } else {
       const notZeroValue: boolean =
         currencyObj?.float !== undefined && currencyObj.float > 0;
-      if (!isFiat && currencyObj && notZeroValue) {
+      if (!isFiat(currency) && currencyObj && notZeroValue) {
         const bchType: string = currencyObj.currency;
         setText(`Send ${currencyObj.string} ${bchType}`);
         query.push(`amount=${currencyObj.string}`);
@@ -380,7 +379,7 @@ export const Widget: React.FC<WidgetProps> = props => {
 
     const goal = getCurrencyObject(cleanGoalAmount, currency);
 
-    if (!isFiat) {
+    if (!isFiat(currency)) {
       if (goal !== undefined && progress.float > 0) {
         setGoalPercent((100 * progress.float) / goal.float);
         const string = progress.string;
--- a/react/src/components/Widget/Widget.tsx
+++ b/react/src/components/Widget/Widget.tsx
@@ -18,13 +18,9 @@ import { formatPrice } from '../../util/format';
 import { Button, animation } from '../Button/Button';
 import BarChart from '../BarChart/BarChart';
 
-import {
-  satoshisToBch,
-  getCurrencyObject,
-  currencyObject,
-} from '../../util/satoshis';
+import { getCurrencyObject, currencyObject } from '../../util/satoshis';
 import { useAddressDetails } from '../../hooks/useAddressDetails';
-import { currency, getSatoshiBalance } from '../../util/api-client';
+import { currency, getAddressBalance } from '../../util/api-client';
 import { randomizeSatoshis } from '../../util/randomizeSats';
 import PencilIcon from '../../assets/edit-pencil';
 
@@ -196,8 +192,8 @@ export const Widget: React.FC<WidgetProps> = props => {
 
     (async (): Promise<void> => {
       if (addressDetails) {
-        const { satoshis } = await getSatoshiBalance(to);
-        setTotalSatsReceived(satoshis);
+        const { balance } = await getAddressBalance(to);
+        setTotalSatsReceived(balance);
       }
     })();
   }, [addressDetails, totalReceived, totalSatsReceived]);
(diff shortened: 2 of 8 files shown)


Klakurka (changes requested):
- Can we change the button below the QR code back to 'Send with XXX wallet'?
- Maybe not let them denominate XEC in BCH or BCH in XEC. So maybe just remove BCH and XEC as options for `currency`?
- When the value input is editable, it really likes forcing a 0 at the front. If a number is entered, let's clear the leading zero. 
image
- It shouldn't be possible to enter more than 2 decimals when editing XEC or more than 8 when editing BCH (could make this a new task).
- hoverText doesn't seem to work. It's still showing 'Send XEC' for me.
-

react/src/util/api-client.ts
```diff
@@ -93,12 +93,26 @@ export default {
...
 
+export function isFiat(unknownString: string): unknownString is fiatCurrency {
+  return fiatCurrencies.includes(unknownString as fiatCurrency);
+}
+
+export function isCrypto(
+  unknownString: string,
+): unknownString is cryptoCurrency {
+  return cryptoCurrencies.includes(unknownString as cryptoCurrency);
+}
```
Gattaca1, line 115: husky absolutely will not let me format this the same as `isFiat`. Not sure whats going on with that

Gattaca1:
> * Can we change the button below the QR code back to 'Send with XXX wallet'?

Yep, I'll address that in this branch

>     * Maybe not let them denominate XEC in BCH or BCH in XEC. So maybe just remove BCH and XEC as options for `currency`?
Yep, I'll address that in this branch

>     * When the value input is editable, it really likes forcing a 0 at the front. If a number is entered, let's clear the leading zero. 
>     * It shouldn't be possible to enter more than 2 decimals when editing XEC or more than 8 when editing BCH (could make this a new task).
Good idea - though since its an existing issue with BCH too, I think its out of scope for this branch

>     * hoverText doesn't seem to work. It's still showing 'Send XEC' for me.
Yep, I'll address that in this branch

Gattaca1:
I'm going to leave this branch in DRAFT until server support for currency pricing is merged. It doesn't feel right to merge this with a hardcoded XEC price

Gattaca1:
>     * Maybe not let them denominate XEC in BCH or BCH in XEC. So maybe just remove BCH and XEC as options for `currency`?

Actually this is a bigger change than I expected. Going to address this in the next stacked branch

react/src/components/Widget/Widget.tsx
```diff
@@ -263,46 +262,50 @@ export const Widget: React.FC<WidgetProps> = props => {
...
 
+    let addressType: currency = 'AUD';
+
+    if (isValidCashAddress(address)) {
+      addressType = 'BCH';
+      setWidgetButtonText(`Send with BCH wallet`);
+    } else if (isValidXecAddress(address)) {
+      addressType = 'XEC';
+      setWidgetButtonText(`Send with XEC wallet`);
+    }
```
ScottMcDermid, line 273: I think you can leverage the `validCurrencyType` function to avoid if statements here.

```js
const currency = validCurrencyType(address);
setWidgetButtonText(`Send with ${currency} wallet`);
```

Klakurka (changes requested):
- It doesn't seem to be polling the server at all beyond just getting the current price.
- If you have a goal value set up, it just spins forever because it never attempts to get the address balance.

react/src/components/Widget/Widget.tsx
```diff
@@ -263,46 +262,50 @@ export const Widget: React.FC<WidgetProps> = props => {
...
+
+      if (bchAmount) {
+        if (isValidCashAddress(address)) {
+          setText(
+            `Send ${currencyObj.string} ${currencyObj.currency} = ${bchAmount.string} BCH`,
+          );
+        } else if (isValidXecAddress(address)) {
+          setText(
+            `Send ${currencyObj.string} ${currencyObj.currency} = ${bchAmount.string} XEC`,
+          );
```
ScottMcDermid, line 288: Looks like you could leverage your `validCurrencyType` function here again to simplify things

```js
const addressCurrency = validCurrencyType(address);
setText(`Send ${currencyObj.string} ${currencyObj.currency} = ${bchAmount.string} ${addressCurrency}`);
```

react/src/components/Widget/Widget.tsx
```diff
@@ -263,46 +262,50 @@ export const Widget: React.FC<WidgetProps> = props => {
...
         url = prefixedAddress + (query.length ? `?${query.join('&')}` : '');
         setUrl(url);
       } else {
         if (isValidCashAddress(address)) {
           setText(`Send any amount of BCH`);
-          setWidgetButtonText(`Send any amount of BCH`);
         } else if (isValidXecAddress(address)) {
           setText(`Send any amount of XEC`);
-          setWidgetButtonText(`Send any amount of XEC`);
         }
```
ScottMcDermid, line 309: Looks like you could leverage your validCurrencyType function here again to simplify things

```js
const addressCurrency = validCurrencyType(address);
setText(`Send any amount of ${addressCurrency}`);
setWidgetButtonText(`Send any amount of ${addressCurrency}`);
```

react/src/components/Widget/WidgetContainer.tsx
```diff
@@ -94,12 +92,18 @@ export const WidgetContainer: React.FC<WidgetContainerProps> = withSnackbar(
...
-        if (isFiat) {
-          const data = await getFiatPrice(currency);
+        if (isFiat(currency) && isValidCashAddress(address)) {
+          const data = await getBchFiatPrice(currency);
+
+          const { price } = data;
+          setLoading(false);
+          setPrice(price);
+        } else if (isFiat(currency) && isValidXecAddress(address)) {
+          // TODO: THIS SHOULD POINT TO THE XEC FUNCTION. THIS IS TEMPORARY WHILE WAITING FOR BACKEND
```
ScottMcDermid, line 104: Is this done? Looks like you're using an XEC function

react/src/components/Widget/WidgetContainer.tsx
```diff
@@ -118,15 +122,15 @@ export const WidgetContainer: React.FC<WidgetContainerProps> = withSnackbar(
       (transaction: any, satoshis: number) => {
         if (sound && !hideToasts) txSound.play().catch(() => {});
 
-        const receivedAmount = satoshisToBch(satoshis);
+        const receivedAmount = satoshis;
 
         if (!hideToasts)
           // TODO: This assumes only bch
```
ScottMcDermid, line 128: Is this out of scope for adding XEC support?

react/src/util/api-client.ts
```diff
@@ -63,16 +77,32 @@ export const getTransactionDetails = async (
...
+
+export const cryptoCurrencies = ['BCH', 'XEC'] as const;
+type cryptoCurrenciesTuple = typeof cryptoCurrencies; // readonly ['BCH', 'XEC']
+export type cryptoCurrency = cryptoCurrenciesTuple[number]; // "BCH" | "XEC"
+
 export type currency = cryptoCurrency | fiatCurrency;
 
+export function isFiat(unknownString: string): unknownString is fiatCurrency {
+  return fiatCurrencies.includes(unknownString as fiatCurrency);
+  // this is a comment
```
ScottMcDermid, line 97: It sure is

react/src/util/satoshis.ts
```diff
@@ -1,22 +1,14 @@
...
 
 const BCHdecimals = 8;
-// const XECdecimals = 2;
-
-export const satoshisToBch = (satoshis: number): number =>
-  +(satoshis / 100_000_000).toFixed(8);
-
-export const bchToSatoshis = (bch: number): number =>
-  Math.round(bch * 100_000_000);
+const XECdecimals = 2;
```
ScottMcDermid, line 6: I'd use the constant naming scheme for this: `XEC_DECIMALS`

ScottMcDermid (changes requested):
Update the `README` to include XEC information wherever BCH is mentioned

react/src/util/api-client.ts
```diff
@@ -32,14 +33,27 @@ export const getUTXOs = async (
   return await res.json();
 };
 
-export const getFiatPrice = async (currency: currency): Promise<PriceData> => {
```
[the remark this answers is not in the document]
Gattaca1: totally - i'm working towards splitting `currency` back into `fiatCurrency` and `cryptoCurrency`. They are treated as types, but there are lots of places where its not appropriate to have some of the values of the type.
eg: `${rootUrl}/price/bitcoincash/${_.lowerCase(currency)}`,
only `cryptoCurrency` should be valid there. Its just so baked in that the only crypto currency is `BCH` that its hard to pick apart


the next branch is getting rid of `currency` altogether, so i'll address this function in that branch

react/src/components/Widget/WidgetContainer.tsx
Gattaca1: oops, forgot to remove

react/src/components/Widget/Widget.tsx
Gattaca1: done

react/src/components/Widget/Widget.tsx
Gattaca1: done

react/src/util/address.ts
```diff
@@ -22,7 +23,18 @@ export const isValidXecAddress = (address: string): boolean => {
   }
 };
 
+export const validCurrencyType = (address: string): currency => {
```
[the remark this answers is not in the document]
Gattaca1: done

react/src/components/Widget/Widget.tsx
Gattaca1: done

react/src/components/Widget/WidgetContainer.tsx
Gattaca1: Removing `sats` and `bits` as currencies was a precursor to adding `xec`. Just fixing bugs related to that as i go since its within the same stacked branch

react/src/util/address.ts
```diff
@@ -23,18 +23,18 @@ export const isValidXecAddress = (address: string): boolean => {
...
 
-export const validCurrencyType = (address: string): currency => {
+export const getCurrencyTypeFromAddress = (address: string): currency => {
   if (isValidCashAddress(address)) {
     return 'BCH';
   } else if (isValidXecAddress(address)) {
     return 'XEC';
   } else {
-    return 'GBP';
+    throw new Error();
```
ScottMcDermid, line 32: Throw in a descriptive string or something for now for the error instance.

ScottMcDermid (changes requested):
Code changes look good. Just one new comment.

react/src/util/address.ts
ScottMcDermid: I don't think this is resolved yet. Add some parameter like `new Error( 'Invalid address' )`

ScottMcDermid (changes requested):

react/src/util/address.ts
Gattaca1: you are looking at an old commit

ScottMcDermid (dismissed):

ScottMcDermid:
Update the `./docs/README.md` file

react/.env
```diff
@@ -1,2 +1,3 @@
 IMAGE_INLINE_SIZE_LIMIT=1000000
-REACT_APP_API_URL=https://api.paybutton.org
\ No newline at end of file
+REACT_APP_API_URL=https://api.paybutton.org
+REACT_APP_API_PRICING_URL=http://localhost:3000
```
Klakurka, line 3: Why is this set to localhost by default and the main URL using api.paybutton.org by default?

Klakurka (changes requested):
As before, I can't get it to poll at all (tested in Storybook).

Klakurka:
Rebase?

react/.env
ScottMcDermid: Let's set this to be the same as the main API host for now. Until we run into scalability issues, it should be fine.

Klakurka (approved):
Looks fine to me.

PR #196, opened 2023-02-02
polling back online

vorg-code:
In the validation, we were requiring that the address should be a valid xec AND cash address at the same time, that makes no sense, I believe we wanted to validate if it was a valid xec OR a valid cash address.


Klakurka (approved):
Ok so looks like one of the two server endpoints isn't compatible but we can fix that in a different branch, likely on the server but we can talk about it.

Issue #151, opened 2022-06-08
Transaction check API call happening twice


Klakurka:
Happens on both the normal button/dialog and the widget.

PR #198, opened 2023-02-05
[Issue #151] fix: transaction api called twice

vorg-code:
Description:

When opening the network tab and opening the qrcode Widget either by:
 1. clicking on any child of the menu on the left `Paybutton` and then clicking on the button on the right
 2. clicking on any child of the menu on the left `Widget`
You can see that, after 3 seconds, requests for the transactions endpoint were made in duplicate before this PR, on this branch this is solved and only 1 request is made each time.

Test plan:

Described above.

We were creating the hook `useAddressDetails` twice, once in the parent `WidgetContainer` and again in the child `Widget`, since we have a page with the child and not the parent (2.) and another one with both (1.), I made an optional parameter when creating the child that allows the child only to have this hook and update the parent (if parent needs it).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 node_modules
 dist
+.env.local
--- a/react/src/components/Widget/Widget.tsx
+++ b/react/src/components/Widget/Widget.tsx
@@ -41,7 +41,6 @@ export interface WidgetProps {
   theme?: ThemeName | Theme;
   foot?: React.ReactNode;
   disabled: boolean;
-  totalReceived?: number | null;
   goalAmount?: number | string | null;
   currency?: currency;
   animation?: animation;
@@ -119,7 +118,6 @@ export const Widget: React.FC<WidgetProps> = props => {
     loading,
     success,
     successText,
-    totalReceived,
     goalAmount,
     ButtonComponent = Button,
     currency = getCurrencyTypeFromAddress(to),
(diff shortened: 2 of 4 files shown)


Klakurka (changes requested):
Actually it doesn't build for me:

> (rpt2 plugin) Error: Documents/git/paybutton/react/src/components/Widget/WidgetContainer.tsx(5,1): semantic error TS6133: 'useAddressDetails' is declared but its value is never read.

vorg-code:
@Klakurka fixed

Klakurka (changes requested):
Now I get this:

> /Documents/git/paybutton/react/src/components/Widget/WidgetContainer.tsx(212,23): semantic error TS2339: Property 'unconfirmedTransactionsList' does not exist on type 'never'.

Klakurka (changes requested):
The issue where it's calling the API twice each time looks fixed.

There's still some (possibly related) strange behaviour when you go to open the payment dialog initially though. It sends out 4 calls to `/address/balance/xxx` in addition to the calls to the other 2. It should just send 1 call to get the balance, then continue polling after.

vorg-code:
@Klakurka indeed, that code was poorly written, it made useEffect run more times than it needed, also we weren't getting the balance correctly in the frontend, I removed some unnecessary code in the widget container (it requested balance from the server, but never used it), also some unnecessary `await`s.

Klakurka (approved):

ScottMcDermid (approved):
Code looks good!

Issue #207, opened 2023-06-06
Mixing HTML + JS buttons on the same page results in only the JS buttons rendering


Klakurka:
This issue can be seen on https://paybutton.org.

Klakurka:
I think this was actually fixed in February and I was just testing on an older build.

Issue #205, opened 2023-05-28
'Goal' not working


Klakurka:
![image](https://github.com/PayButton/paybutton/assets/4713204/bdba101f-d82b-4e5a-bc0d-e3d72e3f8dc1)

`bitcoincash:qpfx6l0x85kc5qaah4wx9nahezcgqefnss2sp0cy06`

This has been broken since switching to our own API.

In the above screenshot, the balance should not be $0.

PR #209, opened 2023-06-08
[Issue #205] fix: Goal price fix

chedieck:
Description:
---
Fixes the `goal` feature, which was broken. Probably related to the obsolete Currency Formatter lib being used. In their own [github page](https://github.com/smirzaei/currency-formatter): 
image

This is still being used in other places. I did not care to refactor it all because this would require more work than the necessary to fix the goal issue, but would be a good thing to do post V1.

Test plan:
---

Compile the `paybutton.js` file locally by running `npm run build` first on the `react/` folder and then on the `paybutton/` folder.

Then, create a HTML file e.g:

```
<!DOCTYPE html>
<html>
<head>
  <title>lkasdfjaskl</title>
  <!--<script src="https://unpkg.com/@paybutton/paybutton/dist/paybutton.js"> -->
  <script src="/home/example-path/client-paybutton/paybutton/dist/paybutton.js">
  </script>

</head>
<body>
  <h1>Botão</h1>
  <div id="botaoo"></div>
  <script>
    var config = {
      to: 'bitcoincash:qpfx6l0x85kc5qaah4wx9nahezcgqefnss2sp0cy06',
      amount: 4.5,
      goalAmount: 1,
      canEdit: false,
      currency: 'BCH',
      text: 'Tip Us a Coffee!',
      hoverText: 'One BCHinno',
      theme: {
        palette: {
          primary: '#ee8b2b',
          secondary: '#fefbf8',
          tertiary: '#504030'
        }
      },
      animation: 'invert',
      successText: 'Purchase Complete!',
      randomSatoshis: true
    };
    PayButton.render(document.getElementById('botaoo'), config);
  </script>
</body>
</html>
```

... and change the line at the top `  <script src="/home/example-path/client-paybutton/paybutton/dist/paybutton.js">` to point to your locally created `paybutton.js` path. Notice the commented line above loads the js from the remote. You can uncomment it and comment the one below to test how it differs.

### DIFF
--- a/react/src/util/format.ts
+++ b/react/src/util/format.ts
@@ -1,9 +1,11 @@
 import currencyFormat from 'currency-formatter';
 import { currency } from './api-client';
 
-export const BCH_DECIMALS = 8;
-export const XEC_DECIMALS = 2;
-export const FIAT_DECIMALS = 2;
+export const DECIMALS = {
+  BCH: 8,
+  XEC: 2,
+  FIAT: 2,
+};
 
 export const amount = (x?: number | null): string | undefined => {
   const clean = +x!;
@@ -17,7 +19,7 @@ export const amount = (x?: number | null): string | undefined => {
 export const formatPrice = (
   price: number,
   currencyType: currency,
-  precision = FIAT_DECIMALS,
+  precision = DECIMALS.FIAT,
 ) => {
   return Intl.NumberFormat(undefined, {
     style: 'currency',
--- a/react/src/util/format.ts
+++ b/react/src/util/format.ts
@@ -1,6 +1,10 @@
 import currencyFormat from 'currency-formatter';
 import { currency } from './api-client';
 
+export const BCH_DECIMALS = 8;
+export const XEC_DECIMALS = 2;
+export const FIAT_DECIMALS = 2;
+
 export const amount = (x?: number | null): string | undefined => {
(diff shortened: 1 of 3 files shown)


react/src/util/format.ts
```diff
@@ -1,6 +1,10 @@
 import currencyFormat from 'currency-formatter';
 import { currency } from './api-client';
 
+export const BCH_DECIMALS = 8;
+export const XEC_DECIMALS = 2;
+export const FIAT_DECIMALS = 2;
+
```
Klakurka, line 7: What do you think about making this into an obj like:

`DECIMALS = { BCH: 8, XEC: 2, FIAT: 2 }`

?

Klakurka (changes requested):
Works perfectly.

Klakurka:
This seems to have also fixed #207... I think I was testing on an old build.

chedieck:
> This seems to have also fixed #207... I think I was testing on an old build.

Yes, I was about to say that. I tested both the remote `paybutton.js` after downloading it to my local system (to isolate this "remote file" variable) and the one generated in this branch, and found the same.

react/src/util/format.ts
chedieck: Done.

Klakurka (approved):

PR #211, opened 2023-06-09
chore: Better compiling config

chedieck:
Description
---
Better config to dev, makes it so that it takes 40s to compile it in my machine instead of 10m.

Test plan
---
Before, running `npm run start` on the `paybutton/` folder should take significantly longer than running (now renamed to `npm run dev`) it now.

 
Remarks
---
Makes so that we don't minimify it and remove comments on the output JS file if running `npm run dev`. This is only useful for production, so we only do it on production.


chedieck:
> What would be the command for prod?

`npm run build`. Should be unaltered.

Klakurka (changes requested):
Can you try merging master into this?

Klakurka (changes requested):
There's some performance issues loading HTML buttons when using this. Anything you want to do about it? The prod npm build is fine.

chedieck:
> Can you try merging master into this?

Done

Klakurka (approved):
OK so the issue is that livereload is timing out after 20s on the server I'm running it on which I don't think is a big deal.

Issue #218, opened 2023-06-27
Success toast/process doesn't fire when correct payment is detected


Klakurka:
When the correct amount is sent to an address for payment, execute the success toast/process. This also includes closing the payment dialog and running the 'success' callback.

Klakurka:
Sound still fires twice.

Klakurka:
Actually now I can't even get the process to fire at all so we'll need to look into that.

Klakurka:
Turns out my issue was an issue related to https://github.com/PayButton/paybutton/issues/223

PR #219, opened 2023-07-01
[Issue #218] fix: toast when payment received

chedieck:
Related to #218

Description
---
Fixes the toast effect.

Test plan
---
Create a button with one of your addresses and send some XEC to it to check if the effect arrives as expected
**Right now there is a problem in the prod server where the server stops registering new txs**. This won't be much testable until this is fixed.

Remarks
---
The client was still expecting the response of the `api/address/transaction/[addr]`  to be formatted as grpc. This goes on to fix the expectation to the new standardized format.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,17 +1,13 @@
-<!-- Insert the related issue below -->
 Related to #
----
-- [ ] Solves
-- [ ] Partially solves
 
-
-<!-- Remove this section if there are no dependencies -->
+<!--
 Depends on
 ---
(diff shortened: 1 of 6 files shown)


Klakurka (changes requested):
- Payment seems to be detected twice when it happens: 
image
- After closing the dialog, opening up the button again has the success process run again -- with the 'Received' toast appearing multiple times: 
image
- Success sound plays when closing hte payment dialog.

Klakurka (approved):
Approved knowing that the issues are being addressed by #220.

Issue #224, opened 2023-07-22
Sucess & Transaction callbacks not firing


Klakurka:
`Success` should fire when the exact amount is received.

`Transaction` should fire when any amount is received.

PR #225, opened 2023-07-25
fix: callbacks for html paybutton

chedieck:
Related to #224

Description
---
Fixes the problem where the `on-success` and `on-transaction` attributes of a HTML paybutton wouldn't work.

Test plan
---
Create and HTML paybutton with said attributes and check if they fire as they should. (`onSuccess` fires when the amount is equal to the expected, `onTransaction` fires regardless of the amount).


Klakurka (approved):
I got it to work once but the server is still having issues so I want to re-test after that's been sorted out as well.

Issue #231, opened 2023-08-13
Setting 'currency=BCH' when in XEC mode simply shows XEC values  + "BCH"


Klakurka:
My suggestion here is to consolidate the 'XEC' and 'BCH' options into one. Possibly 'crypto' ? Which would also continue to be the default value.

![image](https://github.com/PayButton/paybutton/assets/4713204/38e7f6dc-2456-4a75-8bf2-6dd8753740d9)

PR #232, opened 2023-08-14
[Issue #231] fix: currency default value

chedieck:
Related to #231

Description
---
Fixes the behavior of the `currency=` parameter to use the same crypto as the address.

Test plan
---
1 - Not setting `currency=` parameter would show the price in BCH even if it was a XEC address. Now it will should show the amount in the correct currency.
2 - This can also be achieved by setting `currency=crypto`
3 - Setting `currency=BCH` in a XEC addr or vice-versa should have no effect. 

Remarks
---
I changed the `react/` dev build command from `start` to `dev`, so it uses the same logic of production/development with `build`/`dev`  as the `paybutton/` folder.

### DIFF
--- a/react/package.json
+++ b/react/package.json
@@ -18,7 +18,7 @@
     "build": "microbundle --jsx React.createElement --format modern,cjs",
     "lint": "eslint 'src/**/*.{ts,tsx}' --fix",
     "prettier": "prettier --write 'src/**/*.ts'",
-    "start": "concurrently npm:watch npm:storybook",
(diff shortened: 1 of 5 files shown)


chedieck:
> 1. When you have USD set, it's still showing 'XEC'
> 
>     2. When you have the goal amount set, it's showing 'XEC' 3 times (it should only show it once at the end; the bolded one)
> 


Nice catch, fixed and using less code now.

Klakurka (approved):

PR #241, opened 2023-09-01
fix: if no amount on button, use 0

chedieck:
Related to #

Description
---
Fix issue where button with no amounts would lead to opening the cashtab extension with no address.

Test plan
---
With the new paybutton.js buttons that have no amount should open the cashtab extension normally with the button address.


Klakurka (changes requested):
It sets the input to 0 in Cashtab. Surely we can have it just be empty?

chedieck:
> It sets the input to 0 in Cashtab. Surely we can have it just be empty?

Should be fixed now

Klakurka (approved):

Issue #239, opened 2023-08-31
Add option to override the default server


Klakurka:
Be sure to update /docs as well.

PR #242, opened 2023-09-01
[#239] feat: URLs as parameters

chedieck:
Related to #239

Depends on
---
- [ ] #241

Description
---
- Adds option of adding the `ws-base-url` and `api-base-url` parameters to a paybutton (JS would be `wsBaseUrl` and `apiBaseUrl`) to override the defaults.

Test plan
---
With the JS compiled from this branch, create a HTML & JS paybutton overriding this URLs (see the modified docs for more info), it should override as expected.
 
Remarks
---

- Removes the `paybutton-config.json`, uses simple constants in the code to define the default values
- Simplify the existance of `apiBaseURL` and `apiPriceBaseURL` to just `apiBaseUrl`. This `apiBaseUrl` is already used to fetch a lot of different info, all from the same place (our server), so it made no sense to distinguish the price like this.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -671,9 +671,9 @@ disabled = true
 
 <!-- tabs:end -->
 
-## wsBaseURL
+## wsBaseUrl
 
-> **The ‘wsBaseURL’ parameter specifies the URL for the websocket where the information about new transactions should come.
+> **The ‘wsBaseUrl’ parameter specifies the URL for the websocket where the information about new transactions should come.
 
 ?> This parameter is optional. Default value is ‘https://socket.paybutton.org’. Possible values are any string defining an URL.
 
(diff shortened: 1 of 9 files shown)


Klakurka (changes requested):
I think it's a step back to remove the config file... for us it doesn't matter but for somebody who's wanting to run it themselves, they'll need to update the code which is not what we want.

chedieck:
> I think it's a step back to remove the config file... for us it doesn't matter but for somebody who's wanting to run it themselves, they'll need to update the code which is not what we want.

Indeed, reverted it.

Klakurka (changes requested):
Works great -- just a question about commented out code.

chedieck:
> Works great -- just a question about commented out code.


Removed them.

Klakurka (approved):

Issue #247, opened 2023-09-08
'random-satoshis' not working


Klakurka:
Random satoshis aren't working.

Unless BIP 70 is in use, this should be the default.

PR #251, opened 2023-09-11
[#247] fix: random satoshis for both currencies

chedieck:
Related to #247

Description
---
Makes so that the `random-satoshis` option works well for both XEC and BCH, also makes `true` the default for it.

Test plan
---
1. Unless `random-satoshis="false"` is explicitly set, should be true.
2. For BCH and XEC, the last 4 digits should be ignored and randomized. This means the end should vary from `0.00000000` to `0.00009999` with BCH and from `0.00` to `99.99` on XEC.

### DIFF
--- a/paybutton/src/index.tsx
+++ b/paybutton/src/index.tsx
@@ -146,7 +146,7 @@ function findAndRender<T>(className: string, Component: React.ComponentType<any>
       }
 
       props.hideToasts = attributes.hideToasts === 'true';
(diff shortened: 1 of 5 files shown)


chedieck:
> Math isn't right when currency is set to fiat.


Fixed

Klakurka (approved):
I think in the past we actually added the value onto the base amount but I'm totally fine with this approach. It matches what we describe in the docs anyway.

PR #255, opened 2023-09-12
Add !important to styles

johnkuney:
Related to #

Description
---
We noticed on one site our styles were being overridden. By adding !important to the styles it can help prevent this

Test plan
---


Klakurka (approved):
There's still at least one small issue left but we're going to leave it as it seems tough to prevent.

This is at least much better thann before.

PR #268, opened 2023-09-14
[#265] feat: passing currecy object as props

### LINKED ISSUE #265: Success checkmark + faded QR code not working

chedieck:
Related to #265

Description
---
Fixes #265. If the button had `random-satoshis`, the success message + fading of the QR code wouldn't work, now it should work normally.

Test plan
---
Check that any button, upon receiving the expected amount, triggers the onSuccess routine.


chedieck:
This actually does not fixes #265, since #265 was happening for (at least) two different reasons. One is solved here, the random satoshis would break the `onSuccess` check. The other is #273

Klakurka (approved):

PR #269, opened 2023-09-14
feat: onSuccess and onTransaction do not fire both

chedieck:
Depends on
---
- [ ] #268

Description
---
Changes so that the behavior of the `onSuccess` and `onTransaction` functions is:
`onSuccess`: fires if the incoming transaction amount is the same of the button. If no amount is set, never fires.
onTransaction: fires in all other cases

Test plan
---
Check that the behavior described above is correct for buttons with or without amount.

### DIFF
--- a/react/src/components/Widget/Widget.tsx
+++ b/react/src/components/Widget/Widget.tsx
@@ -313,7 +313,7 @@ export const Widget: React.FC<WidgetProps> = props => {
       if (convertedCurrencyObj) {
         thisAmount = convertedCurrencyObj.float
       } else {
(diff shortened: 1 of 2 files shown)


Klakurka:
We'll need to update the docs too.

Klakurka (changes requested):
I can only get Transaction to fire.

chedieck:
> I can only get Transaction to fire.

I forgot to add the other branch to it, updated as a dependency, should work now.

Klakurka (changes requested):
I can still only get on-transaction to fire.

PR #272, opened 2023-09-15
[#266] fix: canEdit option

chedieck:
Related to #266

Depends on
---
- [ ] #269

Description
---
Fixes the `canEdit` option, which was not working.

Test plan
---
After building, create some buttons that have `can-edit="true"`. Then check that the user can edit the amount and also that sending a payment to that button will correctly distinguish the `onSuccess` and `onTransaction` function.


Klakurka (approved):
Looks good, just need to fix on-success never firing in the other branch.

PR #274, opened 2023-09-15
[#273] fix: fiat amount mistakenly checked

### LINKED ISSUE #273: onSuccess not working for fiat buttons
Buttons that have a fiat amount only execute `onTransaction`.

chedieck:
Related to #273

Description
---
Fixes #273 

Test plan
---
Create a button with a fiat `currency` e.g. `currency="CAD"`. Then, after making a payment, make sure `onSuccess` and `onTransaction` are being distinguished.


Klakurka (changes requested):
Three issues I ran into:

> - https://github.com/PayButton/paybutton/issues/275
> - Excessive console logging (I must have missed this in a previous branch) -- there should never be anything sent to the browser console log

Those should be addressed in another PR, as of:

> - Neither on-success or on-transaction functions fire for me

I'll try to reproduce.

Klakurka (approved):
Works for me now so not sure what was going on before.

Issue #290, opened 2023-11-03
Widget CSS getting overridden like the dialog was previously


Klakurka:
Probably just need to make sure that the "!important" styles applied in that original branch also work across to the widget.

PR #291, opened 2023-11-06
Add widget style rule

johnkuney:
Related to #290

Description
---
Fixing the font-size for the widget footer not being specified

Test plan
---
Preview the widget and check the footer link is the right size even if other css rules trying to target it 

for example try adding  <style>
      body a {
        font-size: 44px !important;
        color: red !important;
      }
    </style>

### DIFF
--- a/react/src/components/Widget/Widget.tsx
+++ b/react/src/components/Widget/Widget.tsx
@@ -113,7 +113,7 @@ const useStyles = makeStyles({
   footer: ({ theme }: StyleProps) => ({
     fontSize: '0.6rem !important',
     color: `${theme.palette.tertiary} !important`,
-    fontWeight: 'normal !important',
+    fontWeight: 'normal',
   }),
 });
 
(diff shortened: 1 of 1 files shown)


johnkuney:
Was there anything else besides the footer link being overridden?

Klakurka:
Yes - there were several things so maybe it would be easier to check it by viewing it side-by-side with the dialog open in another window.

Klakurka:
image

Notice the missing text here when in dark mode.

johnkuney:
Okay made a few more tweaks

Klakurka (changes requested):
The transparency on the footer 'Powered by PayButton' text is gone somehow -- on both the widget and on the normal dialog.

johnkuney:
Okay how about now

Klakurka (changes requested):
It's being overridden for me:

image

johnkuney:
for the footer color? hmm that value is what is was set at previously, are you sure you built the latest from this branch? Or possible css cache? These are the styles being added to the footer... 

 footer: () => ({
    fontSize: '0.6rem !important',
    color: '#3f51b566 !important',
    fontWeight: 'normal',
  }),

Klakurka (approved):
Nevermind -- all good.

Issue #296, opened 2023-12-21
Remove hyperlink on "Powered by PayButton.org"


Klakurka:
We don't need the extra traffic and it mostly is just a potential misclick-risk for mobile users.

PR #297, opened 2023-12-29
[#296] Remove powered by link

johnkuney:
Related to #296

Description
---
Remove the powered by paybutton link at the bottom of the widget

Test plan
---
Preview a button and check the link is gone and it looks good

### DIFF
--- a/react/src/components/Widget/Widget.tsx
+++ b/react/src/components/Widget/Widget.tsx
@@ -2,6 +2,7 @@ import {
   Box,
   CircularProgress,
   Fade,
+  Link,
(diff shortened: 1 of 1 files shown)


Klakurka (changes requested):
I'm not sure if I had a change of opinion on this or if I meant this originally but IMO we should keep the text, but just remove the `href` around it.

Maybe bonus points for doing a `no-select` style on it?

johnkuney:
ah that makes more sense, added

Klakurka (approved):

Issue #298, opened 2023-12-29
Add onOpen and onClose functions


Klakurka:
These would trigger when the payment dialog (that has the qr code) opens or closes.

We just need to decide if onClose would fire upon a successful payment or only if the dialog was closed manually.

The widget would not have these functions.

Remember to update the docs page.

PR #299, opened 2023-12-29
feat: onClose and onOpen functions

chedieck:
Related to #298

Creates `onOpen` and `onClose` parameters.

Description
---

Test plan
---

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -41,8 +41,8 @@ function myOpenFunction() {
   console.log('Dialog opened.');
 }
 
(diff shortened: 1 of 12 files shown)


Klakurka (changes requested):
Can you add the payment state of the dialog through the onClose callback function?

chedieck:
> Can you add the payment state of the dialog through the onClose callback function?

I added a parameter `success: boolean` that can be used on the callback to check if the user did or did not payed the desired amount

Klakurka (approved):

PR #304, opened 2024-01-04
feat: on open on close functions

chedieck:
Description
---
Adds arguments to onOpen function: address and amount, correctly set up even if random.

Test plan
---
Build the repo and check that `onOpen` works and that the amount and address match even with `randomSatoshis` on.

### DIFF
--- a/docs/zh-cn/README.md
+++ b/docs/zh-cn/README.md
@@ -37,6 +37,14 @@ function myTransactionFunction(txid, amount) {
   console.log( 'Received: ' + amount );
 }
 
+function myOpenFunction() {
+  console.log('Dialog opened.');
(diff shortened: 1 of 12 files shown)


Klakurka (changes requested):
Small edge case in the amount field:

1. Setup a button with editable="true" and on-open="yourFunc"
2. Have your on-open function console log out `amount` and `to`
3. Open dialog and edit the amount
4. Quickly (quickly) close the dialog and open it again. If you're fast enough, the amount will not have had time to reset and will show as whatever you changed the value to, but it will in fact be reset (so they won't match).

Klakurka (approved):

PR #314, opened 2024-01-09
fix: amount rerendering

chedieck:
Description
---
Fixes bug (visible on ecash.land) where Widget would infinitely rerender.


Klakurka (approved):
Looks good. I'll wait to merge until @johnkuney can verify that it's working as well.

PR #319, opened 2024-01-11
[#317] fix: address change subscription

chedieck:
Related to #317

Description
---
Solves #317 

Test plan
---
Make sure that after changing the address of a widget:
1. The old address, when receiving transactions, doesn't trigger `onSuccess`, `onTransaction`, alert, etc.
2. The new address does.


Klakurka (changes requested):
image

Klakurka (approved):
Works like a charm.

Issue #313, opened 2024-01-08
Update Storybook to use XEC


Klakurka:
Use this for the address: `ecash:qrmm7edwuj4jf7tnvygjyztyy0a0qxvl7quss2vxek`

PR #321, opened 2024-01-11
fix: set xec as default currency

lissavxo:
Related to https://github.com/PayButton/paybutton/issues/313

Description
---
- Changed default currency to be XEC instead of BCH 

Test plan
---

Run story book using this [steps](https://docs.paybutton.org/#/?id=developer-quick-start:~:text=of%20the%20project.-,Developer%20Quick%20Start,-Build) 

Check if the components are set to pay with XEC by default

### DIFF
--- a/react/src/components/PayButton/PayButton.stories.tsx
+++ b/react/src/components/PayButton/PayButton.stories.tsx
@@ -48,14 +48,14 @@ UsdAmount.args = {
   amount: 5,
   currency: 'USD',
   text: '$5',
-  hoverText: 'Pay with BCH',
+  hoverText: 'Pay with XEC',
   animation: 'none',
 };
 
 export const withGoal = Template.bind({});
 withGoal.storyName = 'With Goal';
 withGoal.args = {
-  hoverText: 'Pay with BCH',
+  hoverText: 'Pay with XEC',
   animation: 'none',
   goalAmount: 100,
 };
--- a/react/src/components/PayButton/PayButton.stories.tsx
+++ b/react/src/components/PayButton/PayButton.stories.tsx
@@ -4,7 +4,7 @@ import React from 'react';
 import { ThemeName } from '../../themes';
 import PayButton, { PayButtonProps } from './PayButton';
 
-import { defaultWallets, defaultCurrency } from '../../paybutton-config.json';
+import { defaultCurrency, to } from '../../../.storybook/default-args';
 
 export default {
   title: 'PayButton',
@@ -23,9 +23,7 @@ export default {
     onClose: { table: { disable: true } },
     onTransaction: { table: { disable: true } },
   },
-  args: {
-    to: defaultWallets[defaultCurrency],
-  },
+  args: { to },
   parameters: {
     jest: ['Button'],
   },
@@ -48,14 +46,14 @@ UsdAmount.args = {
   amount: 5,
   currency: 'USD',
   text: '$5',
-  hoverText: 'Pay with XEC',
+  hoverText:`Pay with ${defaultCurrency}`,
   animation: 'none',
 };
 
 export const withGoal = Template.bind({});
 withGoal.storyName = 'With Goal';
 withGoal.args = {
-  hoverText: 'Pay with XEC',
+  hoverText: `Pay with ${defaultCurrency}`,
   animation: 'none',
   goalAmount: 100,
 };
--- a/react/src/components/PayButton/PayButton.stories.tsx
+++ b/react/src/components/PayButton/PayButton.stories.tsx
@@ -4,6 +4,8 @@ import React from 'react';
 import { ThemeName } from '../../themes';
 import PayButton, { PayButtonProps } from './PayButton';
 
+import { defaultWallets, defaultCurrency } from '../../paybutton-config.json';
+
 export default {
   title: 'PayButton',
   component: PayButton,
@@ -22,7 +24,7 @@ export default {
     onTransaction: { table: { disable: true } },
   },
   args: {
-    to: 'bitcoincash:qrmm7edwuj4jf7tnvygjyztyy0a0qxvl7q9ayphulp',
+    to: defaultWallets[defaultCurrency],
   },
   parameters: {
     jest: ['Button'],
--- a/react/src/components/PaymentDialog/PaymentDialog.stories.tsx
+++ b/react/src/components/PaymentDialog/PaymentDialog.stories.tsx
@@ -48,14 +48,14 @@ UsdAmount.args = {
   amount: 5,
   currency: 'USD',
   text: '$5',
-  hoverText: 'Pay with BCH',
+  hoverText: 'Pay with XEC',
   animation: 'none',
 };
 
 export const withGoal = Template.bind({});
 withGoal.storyName = 'With Goal';
 withGoal.args = {
-  hoverText: 'Pay with BCH',
+  hoverText: 'Pay with XEC',
   animation: 'none',
   goalAmount: 100,
 };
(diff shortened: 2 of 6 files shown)


Klakurka (changes requested):
There are a few places where we have 'Pay with BCH' as the text. Can you can those to 'Pay with XEC' ?

lissavxo (commented):
Changed default messages using BCH instead of XAC

Klakurka (dismissed):

Klakurka:
I noticed that there's some 'EUR' currency examples in there but I'm fine leaving them as we'll eventually add EUR exchange rates down the road.

react/src/paybutton-config.json
```diff
@@ -1,4 +1,9 @@
 {
   "wsBaseUrl": "https://socket.paybutton.org",
-  "apiBaseUrl": "https://paybutton.org"
+  "apiBaseUrl": "https://paybutton.org",
+  "defaultCurrency": "xec",
```
chedieck, line 4: Since this is very specific to the storybook, I would make that clear in the name. Otherwise it can be confusing as if it was a default address/currency for the paybutton component. Something like `defaultStorybookCurrency`.

react/src/paybutton-config.json
```diff
@@ -1,4 +1,9 @@
...
+  "defaultWallets": {
```
chedieck, line 5: This should be called `Addresses` instead of `Wallets`, so considering the previous comment `defaultStorybookAddresses`

react/src/components/Widget/Widget.tsx
```diff
@@ -148,7 +148,7 @@ export const Widget: React.FC<WidgetProps> = props => {
   const price = props.price;
   const [url, setUrl] = useState('');
   const [userEditedAmount, setUserEditedAmount] = useState<currencyObject>();
-  const [text, setText] = useState('Send any amount of BCH');
+  const [text, setText] = useState('Send any amount of XEC');
```
chedieck, line 151: Here and in the other places you changed hardcoded `BCH`  to `XEC` we could have `defaultStorybookCurrency?.toUpperCase()`.

chedieck (changes requested):

react/src/components/Widget/Widget.tsx
lissavxo: > Here and in the other places you changed hardcoded `BCH` to `XEC` we could have `defaultStorybookCurrency?.toUpperCase()`.

sure, I will make sure to replace this with a logic to reference the default currency name

react/src/paybutton-config.json
Klakurka: That sounds good to me -- PayButton auto-detects the currency based on the address we pass into it.

chedieck: If it is only the storybook, I don't mind changing the var name later if we need to use it, specially in this case since it would be an easy name to do search-and-replace.

But indeed I see that the defaultCurrency is being used on the Widget component, regardless of the storybook. So I think it makes sense to have defaultCurrency, but the address I think should be defaultStorybookAddress (no need to have Wallet in the name, since we don't use Wallet nomenclature in this repo and in the server it means a different thing).

lissavxo: > If it is only the storybook, I don't mind changing the var name later if we need to use it, specially in this case since it would be an easy name to do search-and-replace.
> 
> But indeed I see that the defaultCurrency is being used on the Widget component, regardless of the storybook. So I think it makes sense to have defaultCurrency, but the address I think should be defaultStorybookAddress (no need to have Wallet in the name, since we don't use Wallet nomenclature in this repo and in the server it means a different thing).

I just changed the logic to be inside the `.storybook` since it will be used only by storybook for testing and showcase 
Also implemented it in a way we calculate the current currency by the crypto address

react/src/components/PayButton/PayButton.stories.tsx
```diff
@@ -63,13 +68,13 @@ withUSDGoalCurrency.storyName = 'With USD Goal';
 withUSDGoalCurrency.args = {
   animation: 'none',
   goalAmount: 500000,
-  currency: 'USD',
+  currency: defaultCurrency,
```
chedieck, line 71: This isn't right: `defaultCurrency` here will evaluate to either `'XEC'` or `'BCH'` (in this case to `'XEC'`,  since `defaultArgs.to` is an `ecash:` address), but here it should be `'USD'`. 

I think you can leave the hardcoded 'USD' string since  the story is specific about it (`withUSDGoalCurrency`)

react/.storybook/schemas.ts
```diff
@@ -0,0 +1,3 @@
+export const defaultArgs = {
```
chedieck, line 1: I like this solution for defining the default storybook variables :+1:

react/src/components/PayButton/PayButton.stories.tsx
```diff
@@ -46,14 +51,14 @@ UsdAmount.args = {
   amount: 5,
   currency: 'USD',
   text: '$5',
-  hoverText: 'Pay with BCH',
+  hoverText: 'Pay with '.concat(defaultCurrency),
```
chedieck, line 54: Any reason you decided to go with `.concat` instead of template strings e.g. `` `Pay with ${defaultCurrency}` ``? I'm fine with the way it is, just out of curiosity

react/src/components/PayButton/PayButton.stories.tsx
```diff
@@ -63,13 +68,13 @@ withUSDGoalCurrency.storyName = 'With USD Goal';
...
+  currency: defaultCurrency,
 };
 
 export const withEURGoalCurrency = Template.bind({});
 withEURGoalCurrency.storyName = 'With EUR Goal';
 withEURGoalCurrency.args = {
   animation: 'none',
   goalAmount: 400000,
-  currency: 'EUR',
+  currency: defaultCurrency,
```
chedieck, line 79: Same thing here and in the other files

react/src/components/Widget/Widget.tsx
```diff
@@ -148,7 +148,7 @@ export const Widget: React.FC<WidgetProps> = props => {
   const price = props.price;
   const [url, setUrl] = useState('');
   const [userEditedAmount, setUserEditedAmount] = useState<currencyObject>();
-  const [text, setText] = useState('Send any amount of BCH');
+  const [text, setText] = useState('Send any amount of'.concat(addressType ?? ''));
```
chedieck, line 151: Missing a space, will result in `Send any amount ofXEC`

chedieck (changes requested):

react/src/components/PaymentDialog/PaymentDialog.stories.tsx
```diff
@@ -4,6 +4,11 @@ import React from 'react';
 import { ThemeName } from '../../themes';
 import PaymentDialog, { PaymentDialogProps } from './PaymentDialog';
 
+import { defaultArgs } from '../../../.storybook/schemas';
+import { getCurrencyTypeFromAddress } from '../../util/address';
+
+const defaultCurrency = getCurrencyTypeFromAddress(defaultArgs.to)
```
chedieck, line 10: This could actually be defined just as it is but exported in the schemas file to avoid the repetition.

chedieck (changes requested):
Just left one minor request, other than that LGTM

react/src/components/PaymentDialog/PaymentDialog.stories.tsx
chedieck: NVM would have to import from utils which is not clean anyways

chedieck (approved):

react/src/components/PaymentDialog/PaymentDialog.stories.tsx
lissavxo: Nice catch!
I just improved the logic a little bit.

chedieck (approved):
Looks great

Klakurka (approved):

Issue #302, opened 2024-01-04
Add basic support for OP_RETURN


Klakurka:
- Add a new `op-return` parameter which will add `opreturn=[hex of their string]`
- It needs to be converted to hex so we'll need to do that before adding it into the URI + QR code

Probably helpful: https://github.com/Bitcoin-ABC/bitcoin-abc/blob/master/cashtab/src/opreturn/index.js

We get to decide how we want the OP_RETURN to be decoded.

Klakurka:
Just allowing them to enter it as JSON might be the play but we can talk about it

Either way, we're limited to 223 bytes or less.

Klakurka:
We'll need some sort of error handling if they enter something too long.

Klakurka:
Another format suggestion:

- allow entering a string that contains spaces and colons. Split on spaces first.
- split on colons, where the left side is the key and the right side is the value. If no colon exists then it's just a value.

Klakurka:
Some examples using the above formatting suggestion:

Code: `opreturn="ecash:ABC123XYZ INVOICE#|778 Notes|'I really like eCash!!!!'"`

What would be entered into the database:

1. `ecash:ABC123XYZ`
2. `INVOICE#` = `778`
3. `Notes` = `I really like eCash!!!!`

Klakurka:
https://github.com/Bitcoin-ABC/bitcoin-abc/blob/master/doc/standards/bip21-ecash-additions.md

ethanmackie:
> Some examples using the above formatting suggestion:
> 
> Code: `opreturn="ecash:ABC123XYZ INVOICE#|778 Notes|'I really like eCash!!!!'"`
> 
> What would be entered into the database:
> 
> 1. `ecash:ABC123XYZ` (key would be `null`)
> 2. `INVOICE#` = `778`
> 3. `Notes` = `I really like eCash!!!!`

FYI - the space for notes may be much less than 223 bytes.

i.e. breaking down the OP_RETURN output script:

- 1 byte of [opreturn prefix] + 
- 4 bytes of [protocol identifier] (see [opreturn prefix std](https://github.com/Bitcoin-ABC/bitcoin-abc/blob/master/doc/standards/op_return-prefix-guideline.md))+ 
- 32 bytes of [ecash address] + 
- 14 bytes of [invoice #] (assuming capped at 1M)+ 
- 172 bytes of [Notes] (or approx 86 characters)

[...]

PR #328, opened 2024-01-13
[#302] feat: op return parameter

chedieck:
Related to #302 

Description
---
Allows for a `op-return` parameter.

Test plan
---

This parameter will be prefixed by `paybutton$` and then embedded in the copy + paste URL — CashTab still not supported.

Then, when pasted into ElectrumABC, this should show something like `paybutton$the-data-you-inputted-in-the-op-return-props-attribute`. 

 
Remarks
---

Parsing of this data will be implemented in a next PR

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -635,6 +635,35 @@ onClose = successCallback
 
 <!-- tabs:end -->
 
+## op-return
+
+> **The ‘op-return’ parameter specifies the custom message that will be send with the transaction.
+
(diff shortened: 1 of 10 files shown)


Klakurka (approved):
From here, we'll want to get PayButton-Server passing this back so we can compare it against what is expected -- so we can begin the process of deprecating the random-satoshis feature.

Cashtab support should be just around the corner.

Also note that Cashtab might still be going with `opreturn` for the URIs -- still TBD.

Issue #322, opened 2024-01-11
Update react 16 -> 18.2


chedieck:
NextJS requires React > 18; so our current version lock conflicts with it, creating a problem to apps that use NextJS and want to use paybutton.

PR #334, opened 2024-01-16
Change to yarn

lissavxo:
Related to #322

Description
---

In this PR we changed the package manager to `yarn` and also performed some package  version updates

Test plan
---
Run story book using this steps from the [docs](https://docs.paybutton.org/#/?id=developer-quick-start:~:text=of%20the%20project.-,Developer%20Quick%20Start,-Build)

> ```
> docker-compose up
> Open http://localhost:6006.
> 
> ```

Check if storybook is working as expected


Klakurka (changes requested):
- This would go under 'chore', not 'feat' as it's not a user-facing feature (for next time).
- Can you rebase with master? The BCH -> XEC storybook changes aren't in this branch.
- Running the normal build command (using npm) doesn't work in this branch. Is there a different command for that now?
image

PR #335, opened 2024-01-16
chore: change to yarn

lissavxo:
Related to #322

Description
---

In this PR we changed the package manager to `yarn` and also performed some package  version updates

Test plan
---
Run story book using this steps from the [docs](https://docs.paybutton.org/#/?id=developer-quick-start:~:text=of%20the%20project.-,Developer%20Quick%20Start,-Build)

> ```
> docker-compose up
> Open http://localhost:6006.
> 
> ```

Check if storybook is working as expected


Klakurka:
What's the new build command? We'll need to update the docs -- although the current `cd react && npm i && npm run build && cd ../paybutton && npm i && npm run build` is actually working fine for me right now.

chedieck (changes requested):
Works for me, but I see that there are some files that still refer to `npm`:
react/README.md
react/Dockerfile
docs/README.md

@EDIT 
NVM, actually I forgot to run `yarn` to update the node_modules. Only `yarn build` worked, but `yarn` is not working.

Klakurka:
@lissavxo A few merge conflicts to fix.

Klakurka (changes requested):
I figured that I'd have trouble running `yarn build` -- on Windows, I get: `bash: yarn: command not found`. This is why we had the full multi-step query in the docs.

Any ideas how I can get it working?

Klakurka (changes requested):
- Worth looking at these circular dependencies? 
image
- Doesn't look like `renderWidget` (at least called by vanilla JS) is working: 
image (you can check it here, clicking the "Switch up widget" buttons at the bottom: https://klakurka.com/paybutton/)

lissavxo:
> * Worth looking at these circular dependencies?
> 
> image
> * Doesn't look like `renderWidget` (at least called by vanilla JS) is working:
>   image (you can check it here, clicking the "Switch up widget" buttons at the bottom: https://klakurka.com/paybutton/)
> * Also button text isn't working (again, viewable on my test page). They all say "Donate" even though most of them have custom text set.

-Submitted a version to fix 

and About the Circular dependencies warning, it is something that was already present in the build, this is something related to the sub-dependencies(readeble-streams) of these two libs: `xecaddrjs` and `rollup-plugin-node-polyfills`. Those libs do not receive development at least for 2 years I think it might be the reason we are getting these warnings.

lissavxo (commented):
fixed issues

Klakurka (changes requested):
- 
image
 (I get it on load here: https://klakurka.com/paybutton/ -- one of the buttons isn't generating, the one that is done via .render())

lissavxo (commented):
Fixed rollup

lissavxo (commented):
Fixed

lissavxo (commented):
fixed

Klakurka (changes requested):
I'm still getting the same error as before (PayButton.render is not a function).

Can you check it again?

Klakurka (dismissed):

Klakurka:
Could maybe add some tests but I'll approve it as is. @chedieck can chime in.

docs/README.md
```diff
@@ -848,7 +848,7 @@ PayButton is a community-driven open-source initiative. Contributions from the c
 ### Build
 
 1. `git clone https://github.com/PayButton/paybutton.git`
-2. `cd paybutton/react && npm i && npm run build && cd ../paybutton && npm i && npm run build`
+2. `cd paybutton && yarn build`
```
chedieck, line 851: Much cleaner  :+1:

docs/zh-tw/README.md
```diff
@@ -847,25 +847,22 @@ PayButton是一個社區主導的開放源代碼促進會。此項目的成功
...
 1. `git clone https://github.com/PayButton/paybutton.git`
-2. `cd paybutton/react && npm i && npm run build && cd ../paybutton && npm i && npm run build`
+2. `yarn build`
 
 ### 启动Storybook
 
-1. `cd react && npm i`
-2. `npm start`
+1. `docker-compose up`
 3. 如果網站沒有自動打開，它將在localhost:6006上運行。
```
chedieck, line 855: Should change here: 3. should be 2.

package.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "paybutton",
+  "version": "2.1.2",
+  "description": "The easiest way to accept eCash online",
+  "author": "poldridge",
+  "main": "index.js",
+  "repository": "https://github.com/PayButton/paybutton.git",
+  "license": "MIT",
+  "scripts": {
+    "build:react":"cd react && yarn && yarn build",
```
chedieck, line 10: Could you add `|| cd ..` to all these build commands that `cd` into somewhere? This will make it go back in case any command fails (yarn build or yarn)

package.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "paybutton",
+  "version": "2.1.2",
+  "description": "The easiest way to accept eCash online",
+  "author": "poldridge",
+  "main": "index.js",
+  "repository": "https://github.com/PayButton/paybutton.git",
+  "license": "MIT",
+  "scripts": {
```
chedieck, line 9: Also,  would be good to have scripts that don't run `yarn` besides the ones that run it.

When developing, most of the time we won't have updated/added/removed any modules, so we don't want to be running `yarn`, which takes a considerable time to run, together with `yarn build`.

package.json
```diff
@@ -0,0 +1,21 @@
...
+    "build:react":"cd react && yarn && yarn build",
+    "build:paybutton": "yarn clean:paybutton && cd paybutton && yarn && yarn build",
```
chedieck, line 11: Why `clean:paybutton` but not `clean:react`? I don't think any should be here, is not always we will want to recompile all packages.

package.json
chedieck, line 1: An idea regarding the comments I made here:
1. `build:x` update the node_modules and build it;
3.`hard:build:x` can be the one that cleans node_modules, dist, and redo it all again
3. `soft:build:x` only builds.


Where `x` can be  `react`  `paybutton` or nothing (for both)

react/Dockerfile
```diff
@@ -3,8 +3,8 @@ FROM node:16-alpine
 WORKDIR /app
 
 COPY package*.json ./
-RUN npm install && npm run build --ignore-scripts
+RUN yarn install && yarn run build --ignore-scripts
```
chedieck, line 6: This is wrong, should be `yarn build` not `yarn run build`

react/src/components/PayButton/PayButton.tsx
```diff
@@ -9,15 +9,15 @@ import { currencyObject, getCurrencyObject } from '../../util/satoshis';
 import BigNumber from 'bignumber.js';
 
 export interface PayButtonProps extends ButtonProps {
-  to: string;
+  to?: string | undefined;
```
chedieck, line 12: This one should not be optional. It's the address the payment will go to, so every paybutton/widget will always have one.

chedieck (changes requested):
Left some remarks.

Also, I see that this branch is a bit behind master, since most of the changes there that will lead to conflicst are mine I will take the liberty of merging master into here to make sure we get the right changes.

package.json
lissavxo: if the script fails for some reason it will come back to the root path automatically

package.json
lissavxo: because paybutton takes paybutton/react package as a dependecy so It is reasonable to clean the decency after re-installing it

package.json
lissavxo: sure, let's make the current `yarn build` as a hard version and I will add another command `yarn build:soft` so it runs the build without re-installing the pachages.

package.json
chedieck: Indeed, I forgot it runs in a subshell so no need to worry about that anymore

package.json
chedieck: Hmm, not sure if I understand but `paybutton/react`, being a file dependency, will go into `node_modules/` as a symbolic link, meaning it will always be a pointer to the up-to-date `../react/` directory

package.json
chedieck: Would be good to have an itermediate version too; my notebook takes too long to reinstall `node_modules/` from 0, so I prefer to try simply running `yarn` only first if I have to build with some module update.

lissavxo: if you need to run it with yarn install you run `yarn build`, if you need to run it without it you use `yarn build:soft` 
not sure If I get the third option you are missing

chedieck: Right now, `yarn build`  will execute something like:
`$(cd react && yarn && yarn build) && $(rm -rf paybutton/node_modules paybutton/dist) && (cd paybutton && yarn && yarn build)`

... which means it will remove `paybutton/node_modules` & `paybutton/dist` (but not `react/node_modules` or `react/dist`, which I think is inconsistent)

The three options I am talking about are:
1. Only run the build commands in both directories. Useful for when we know there are no modules changes.
2. Run the build commands after `yarn` in both directories.  Useful for when we know there are modules changes.
3. Run the build commands after removing stuff (`dist/` and `node_modules`) and running `yarn` in both directories.  Useful for when we know there are more significant modules changes or something seems wrong with the current state of node_modules. This one takes longer to run, so I would use more in a situation where I want to be very strict or am trying to solve some issue. 

My request is that we have a clear way to run each one of them, since we're implementing this here.

This could always be done in a separate PR that depends on this one if you want.

package.json
chedieck: Indeed, you're right.

I just found out that `yarn` and `npm` treat differently `file:` dependency resolutions. `npm` would do as I said and create a symbolic link for it; meanwhile yarn actually copies it into `node_modules/`, so changes are not reflected immediately.

That would explain also another issue (I was about to write about) where in my notebook `yarn` takes way  too long to fetch the last package, probably because it's copying it and therefore copying `react/node_modules` itself into `paybutton/node_modules/@paybutton/react/`; a lot of very small files.

Searching about that I found out that this is simply solved by using `link:` instead of `file:` in the `paybutton/package.json`, so unless you see any problem with that I think is the way to go.

chedieck (changes requested):

package.json
lissavxo: we already have the three commands you are referring to 

we do have a separated command to clean the cache before running the build, and this one works together with installing dependencies wich is `clean:build`

the only thing that made be clean the paybutton cache before rebuilding and runing the project it is because it takes react as a dependency so it is reasonable to clean it everytime we run a new build in react folder

chedieck: I see now you added `build:soft`, so indeed it looks good to me. It would be

1. `yarn build:soft`
2. `yarn build`
3. `yarn clean:build`

chedieck (approved):

Klakurka (approved):

PR #339, opened 2024-01-19
[#32] chore: add jest to project

chedieck:
Related to #32

Description
---
Adds `jest` test suite to the project & a sample test.

Test plan
---
`npm run test` inside `react/`

Remarks
---

I had to update typescript in order for it to work


chedieck:
> This error is on `npm i`.

Indeed this is happening right now, you can bypass it using the --legacy-peer-deps on the `npm i`. 
Lissa is currently fixing all these dependency issues right now tho

Klakurka (approved):

PR #340, opened 2024-01-19
[#302] test: op return specs tests

chedieck:
Related to #32, #302 

Depends on
---
- [x] #332 
- [x] #339 

Description
---
Adds tests for the opReturn specs created at #332.  

Test plan
---
First make sure that inside  `react/`  `npm run test` runs fine; and then check the implemented tests to see if there is an edge case missing or some test with wrong expection


Klakurka (dismissed):
Passes my QA (I'm including other manual testing I've done as well) but this is a bit beyond what I'm comfortable doing a code review for so I'll call for backup before approving this.

chedieck:
> Passes my QA (I'm including other manual testing I've done as well) but this is a bit beyond what I'm comfortable doing a code review for so I'll call for backup before approving (and merging) this.

Indeed right now is too big changes because it includes the diff for the other two PRs; one is merged and I've just merged master into this branch; so as soon as the other PR also gets merged it will be reviewable,

react/src/tests/opReturn.test.ts
```diff
@@ -1,7 +1,144 @@
-import { stringToHex } from '../util/opReturn'
+import { exportedForTesting, parseOpReturnProps } from '../util/opReturn'
 
-describe('Sample', () => {
-  it ('', () => {
+const { stringToHex, prependNonceWithPushdata, generatePushdataPrefixedNonce, getDataPushdata } = exportedForTesting
+
+describe('stringToHex', () => {
+  it ('Simple conversion 1', () => {
```
BytesOfMan, line 6: `it ('converts the PayButton lokad id from ascii to hex', () => {`

react/src/util/opReturn.ts
```diff
@@ -1,20 +1,109 @@
...
+  const pushdata = bytesQuantity.toString(16).padStart(2, '0');
+  return `${pushdata}${hexString}`;
 }
 
-export function parseOpReturn(
-  opReturn: string | undefined,
-): string | undefined {
-  if (opReturn === undefined) return undefined;
-  const bytesQuantity = new Blob([opReturn]).size;
+function stringToHex(str: string): string {
```
BytesOfMan, line 37: should decide on a spec for what kind of data will be in each push, then just always use that -- i.e. always ascii, always utf8, always hex to decimal number.

react/src/tests/opReturn.test.ts
chedieck: Hmm that's not what it does tho.  I am assuming lokad is same as pushdata; but this function is used on the user data itself. It's also a generic function;  could be used to encode anything from string to hex using UTF8; not limited to ASCII

BytesOfMan: > I am assuming lokad is same as pushdata

The lokad here is 50415900

> It's also a generic function; could be used to encode anything from string to hex using UTF8; not limited to ASCII

Right -- so you want unit tests named after various things that the function can do.

This particular unit  test is converting the PayButton lokad id from ascii to hex.

'Simple conversion 1' doesn't give any information to an external viewer of this codebase.

BytesOfMan (changes requested):
imo there is a need for an encoding standard. it's not really a feature that it can encode ascii or utf8. want to pick the one you need and put that in the spec.

the better defined the spec is, and the tighter the unit tests are here, the easier and faster it is to process all of these txs on the blockchain to control consistent app functionality.

react/src/tests/opReturn.test.ts
```diff
@@ -1,7 +1,144 @@
...
     expect(stringToHex('PAY\x00')).toBe('50415900')
   })
+  it ('Simple conversion 2', () => {
+    expect(stringToHex('aA1=-*&]]^!\x04\x14')).toBe('6141313d2d2a265d5d5e210414')
+  })
+  it('Works for non ASCII chars', () => {
+    expect(stringToHex('a©æ😂👌')).toBe('61c2a9c3a6f09f9882f09f918c')
```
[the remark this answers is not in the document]
chedieck: I'll respond and address all the other remarks later; but for now I just wanted to clarify this about the encoding that  is very relevant to all of the PR

I have mimicked the way electrumABC does; which is UTF8. UTF8 encodes ASCII chars the same way ASCII does (1-byte; the same byte); so for those there are no difference. as for others; such as ©; which is ASCII extended; there are differences. `a9` is the hex code for it on extended ASCII; but UTF8 prepends with the c2 byte. The emojis ofc require more bytes (4). The idea here is to make sure that these 3 classes of characters all get encoded as expected.

I'll change the name of this test (and others) later to better explain that.

react/src/tests/opReturn.test.ts
```diff
@@ -1,7 +1,144 @@
-import { stringToHex } from '../util/opReturn'
+import { exportedForTesting, parseOpReturnProps } from '../util/opReturn'
 
-describe('Sample', () => {
-  it ('', () => {
+const { stringToHex, prependNonceWithPushdata, generatePushdataPrefixedNonce, getDataPushdata } = exportedForTesting
+
+describe('stringToHex', () => {
+  it ('Simple conversion 1', () => {
     expect(stringToHex('PAY\x00')).toBe('50415900')
```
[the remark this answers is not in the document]
chedieck: `PAY0` is different from `PAY\x00`, but both are ASCII. The `\x00` char is just the NULL char in the ASCII table, encoded to `00` in hex; while `0` would be the literal char 0, encoded to `30` in hex.

Bearing in mind that this is UTF8 conversion, but that UTF8 encoding & ASCII encoding are equal for all ASCII chars.

react/src/util/opReturn.ts
```diff
@@ -1,20 +1,109 @@
...
+// = 205 available bytes
+export const USER_DATA_BYTES_LIMIT = 205;
+
+// Pushdata is self-describing up to 75 bytes, since 0x4c (76 in hex) is
+// a special OP code.
+const SINGLE_PUSHDATA_BYTE_LIMIT = 75;
+
+function prependNonceWithPushdata(hexString: string): string {
+  // 2 hex chars == 1 byte
+  const bytesQuantity = hexString.length / 2;
```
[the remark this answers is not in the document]
chedieck: done

react/src/util/opReturn.ts
chedieck: Commented on top of the function

chedieck:
> imo there is a need for an encoding standard. it's not really a feature that it can encode ascii or utf8. want to pick the one you need and put that in the spec.
> 
> the better defined the spec is, and the tighter the unit tests are here, the easier and faster it is to process all of these txs on the blockchain to control consistent app functionality.

This would be UTF8. I've added a comment on the `stringToHex` function (the one that encodes the user data)

BytesOfMan:
Could you add some tests that show fully completed PayButton OP_RETURN outputs, i.e. what we expect to see as a param value corresponding to the `op_return_raw` key?

In this way we can make sure the spec defined [here](https://github.com/Bitcoin-ABC/bitcoin-abc/blob/master/doc/standards/paybutton.md) agrees with this codebase

BytesOfMan:
https://reviews.bitcoinabc.org/D15223

chedieck:
> Could you add some tests that show fully completed PayButton OP_RETURN outputs, i.e. what we expect to see as a param value corresponding to the `op_return_raw` key?


I think that is what the tests defined in `describe('parseOpReturnProps')` do, isn't it?

In any case I noticed those tests names were not clear and explanatory so I've just changed them.

I thought about also adding two tests for the same examples defined in the specs but:

1. First example has no nonce so can only be added in #341, which implements the nonce as optional
2. Second example could be added; but it's just the same as the 1st + nonce; which would be matched against a regex pattern, which is already being done.

So I think I'll add the test for the exact example 1. on #341; linking to the specs on a comment above the test description.

BytesOfMan:
Ok just doing some checks between this in the spec. Spec right now is

4 pushes
1. Lokad ID
2. 00 ("OP_0")
3. utf8-encoded data
4. utf8-encoded nonce (may or may not appear)

Will the utf-8-encoded nonce always be the same size? If so I will add this to the spec.

What are the inputs and outputs here? Input --> just a PayButton tx? So the data is data about the tx, or about the app, this type of thing? And the nonce is just random?

chedieck:
> Will the utf-8-encoded nonce always be the same size? If so I will add this to the spec.

The nonce is not UTF-8 encoded. Those are random bytes encoded to hex; their usefulness is already materialized in the hex and they are not meant to be decoded.
That said, it is meant to be the same size (8 bytes) for now (we may change that in the future if it presents as a demand for users)

> What are the inputs and outputs here? Input --> just a PayButton tx? So the data is data about the tx, or about the app, this type of thing? And the nonce is just random?

The input is data an user can put when creating a Paybutton by using the `op-return` props, e.g: let's say our PayButton user sells fruits in his website, he may have a PayButton there with props `op-return="product=apple"`. Then, if a buyer uses the PayButton to make the payment, the paybutton-server will see it, parse it, and broadcast a POST request to the user chosen website such that it will arrive to him in the request body: `{data: {product: "apple"}, paymentId: "0dab8f6a6633b0"}` (among with other stuff related to the tx if he wants, e.g. amount, timestamp, etc.)

Then he can use the `{product: "apple"}` data for organizing internal stuff and the paymentId to clear that THAT exact user who made the payment should have an apple (if other users make the same payment, the same amount to the same address, it will have a different paymentId)

BytesOfMan:
is there any case here where the data does not appear but the nonce does? or nothing appears but the lokad signifying it's a paybutton tx?

Klakurka:
If the user doesn't enter a OP_RETURN value and also has `disable-payment-id="true"` then it would just be the lokad id.

Klakurka:
@chedieck conflicts.

chedieck:
> @chedieck conflicts.

Solved

react/src/tests/opReturn.test.ts
```diff
@@ -3,33 +3,38 @@ import { exportedForTesting, parseOpReturnProps } from '../util/opReturn'
...
   })
 });
 
 describe('prependPaymentIdWithPushdata', () => {
-  it ('Simple paymentId prepend 1', () => {
+  it ('Prepends three bytes', () => {
     expect(prependPaymentIdWithPushdata('aa00bb')).toBe('03aa00bb')
   })
-  it ('Simple paymentId prepend 2', () => {
+  it ('Prepends 45 bytes', () => {
```
Klakurka, line 21: Actually I'd like to see a comment or something above to try explaining the "why" on these (including the ones below) as 3, 45, 75, etc all magic numbers to me. I see the 75 one explained lower down but not sure about the others.

Klakurka (changes requested):

BytesOfMan:
> If the user doesn't enter a OP_RETURN value and also has `disable-payment-id="true"` then it would just be the lokad id.
> 
> If the user doesn't enter a OP_RETURN value and also has `disable-payment-id="false" (the default)` then it would be the lokad id + nonce (payment ID). The data would be blank.

Ok. In this case, the spec should be changed so that the data push is also optional. This confusing since the way the spec is now, we know the 1st push after the OP_0 version push is data -- as it is not optional in the spec. If they are both optional, it's possible that the first push after the lokad id is actually the nonce. We need some way of distinguishing between them in the spec.

Different options for handling this.

1. Change the spec so that the nonce can only be present if there is user data
2. Change the spec so that the nonce is not optional and, instead, the data is optional -- every paybutton tx gets a nonce. In this case the nonce should be the 1st push after OP_0 and data should be the 2nd push.
3. We also need to update the spec to give some sort of definition to a PayButton tx that is only the lokad id and no other pushes. It might be better to just have no OP_RETURN msg in this case. Easy fix if you just always have a nonce -- then every one gets a nonce, it's a valid version 0 paybutton tx.

Anyone could spoof these txs but it would be a little harder to spoof txs with a nonce according to your standards than just the lokad id.

chedieck:
>  it's possible that the first push after the lokad id is actually the nonce. We need some way of distinguishing between them in the spec.

Yes, I was thinking about this two. I think the best way to go is to always have the pushdatas, even if `00` so we can always expect these bytes and still leave both optional.

Klakurka:
> > it's possible that the first push after the lokad id is actually the nonce. We need some way of distinguishing between them in the spec.
> 
> Yes, I was thinking about this too. I think the best way to go is to always have the pushdatas, even if `00` so we can always expect these bytes and still leave both optional.

I'm ok with this if we want to go w/ that.

chedieck:
> I'm ok with this if we want to go w/ that.

Alright, should be done.

Klakurka (dismissed):

react/src/tests/opReturn.test.ts
```diff
@@ -85,94 +107,109 @@ describe('getDataPushdata', () => {
...
+    expect(resultPaymentId).toMatch(paymentIdRegex);
+  });
   it('Simple parse OpReturn 2 no paymentId', () => {
-    const fullResult = parseOpReturnProps('my=Longer more=sensible|user|data', true) // 33 bytes
-    expect(fullResult).toBe(allResultsPrefix +
-      '21' + // 33 in hex
-      '6d793d4c6f6e676572206d6f72653d73656e7369626c657c757365727c64617461')
-  })
+    const fullResult = parseOpReturnProps(
+      'my=Longer more=sensible|user|data',
```
BytesOfMan, line 183: Is this supposed to represent a JSON? Does PayButton have an internal way of consistently encoding / decoding this stuff?

react/src/util/opReturn.ts
```diff
@@ -85,21 +98,24 @@ function getDataPushdata(data: string, disablePaymentId=false) {
 
 export function parseOpReturnProps(
```
BytesOfMan, line 99: what does this function do? Looks like it returns a hex string that is ready for `op_return_raw` param?

would help to have some comments about the function's inputs, outputs, mb ref the spec. `parseOpReturnProps` is confusing since it does not seem to be doing this. 

mb `encodePaybuttonOpReturnRaw` or similar. Or mb I am misreading it. 

It's unclear.

react/src/tests/opReturn.test.ts
Klakurka: Yeah this will be on the docs page once that's updated.

If we can decode it with our scheme here, or if they just enter straight JSON (server PR is coming for that one) then it'll save it as JSON. If it can't decode it then it'll just go in as a string.

react/src/util/opReturn.ts
```diff
@@ -55,10 +68,10 @@ function generatePushdataPrefixedPaymentId(bytesAmount: number): string {
   return prependPaymentIdWithPushdata(hexString);
 }
 
-function getDataPushdata(data: string, disablePaymentId=false) {
+function getDataPushdata(data: string, disablePaymentId = false) {
   const bytesQuantity = new Blob([data]).size;
-  // If paymentId is expected, limit is 9 bytes smaller
-  const bytesLimit = USER_DATA_BYTES_LIMIT - (disablePaymentId ? 0 : 9)
+  // If nonce is expected, limit is 8 bytes smaller
```
Klakurka, line 73: Intentionally changed this from `paymentID` to `nonce`?

react/src/util/opReturn.ts
chedieck: It encodes the `op-return=""` props the user has passed to the paybutton.

Indeed `encode` here is better than `parse`; so I've changed to `encodeOpReturnProps` which refers to the function input; the `op_return_raw` ready hex is the output, which is explained in the comments above the function declaration (also added one more comment at the end). Let me know what you think

react/src/util/opReturn.ts
chedieck: Indeed this was a mistake, I've just changed (here and where else it had 'nonce')

Klakurka (approved):

react/src/tests/opReturn.test.ts
```diff
@@ -85,94 +107,109 @@ describe('getDataPushdata', () => {
...
-    expect(resultPaymentId).toMatch(paymentIdRegex)
-  })
+  const allResultsPrefix = '04' + '50415900' + '00';
+  const paymentIdDigits = 18;
+  const paymentIdRegex = /^[0-9a-fA-F]{18}$/;
+  it('Parses undefined data', () => {
+    const fullResult = encodeOpReturnProps(undefined);
+    const resultPaymentId = fullResult?.slice(-paymentIdDigits);
+    const resultNoPaymentId = fullResult?.slice(0, -paymentIdDigits);
+    expect(resultNoPaymentId).toBe(allResultsPrefix + '00'); // 00 is pushdata for no data
```
BytesOfMan, line 130: Shouldn't this be `allResultsPrefix + '00' + '00'` now, per the latest spec? There is no data and no nonce. But the fields are still required.

react/src/tests/opReturn.test.ts
```diff
@@ -85,94 +107,109 @@ describe('getDataPushdata', () => {
...
-    const fullResult = parseOpReturnProps(undefined)
-    const resultPaymentId = fullResult?.slice(-paymentIdDigits)
-    const resultNoPaymentId = fullResult?.slice(0, -paymentIdDigits)
-    expect(resultNoPaymentId).toBe(allResultsPrefix + '00') // 00 is pushdata for no data
-    expect(resultPaymentId).toMatch(paymentIdRegex)
-  })
+  const allResultsPrefix = '04' + '50415900' + '00';
+  const paymentIdDigits = 18;
+  const paymentIdRegex = /^[0-9a-fA-F]{18}$/;
+  it('Parses undefined data', () => {
```
BytesOfMan, line 126: `Encodes undefined data`

react/src/tests/opReturn.test.ts
```diff
@@ -85,94 +107,109 @@ describe('getDataPushdata', () => {
...
+    expect(resultPaymentId).toMatch(paymentIdRegex);
+  });
   it('Empty data no paymentId', () => {
-    const fullResult = parseOpReturnProps('', true)
-    expect(fullResult).toBe(allResultsPrefix + '00') // 00 is pushdata for no data
-  })
+    const fullResult = encodeOpReturnProps('', true);
+    expect(fullResult).toBe(allResultsPrefix + '00' + '00'); // the 00s are pushdata for the data and paymentId respectively
+  });
   it('Simple parse OpReturn 1', () => {
```
BytesOfMan, line 148: Need to move away from these numbered test names. The value of the unit tests is not in the quantity of tests, but in covering every expected case.

If there are 19 cases, we need 19 tests. If there are 5 cases, 5 tests is great.

Here we are testing `it('Encodes a PayButton v0 protocol transaction with custom user data and nonce')`

react/src/tests/opReturn.test.ts
```diff
@@ -85,94 +107,109 @@ describe('getDataPushdata', () => {
...
+    const resultNoPaymentId = fullResult?.slice(0, -paymentIdDigits);
+    expect(resultNoPaymentId).toBe(
+      allResultsPrefix +
+        '10' + // 16 in hex
+        '6d79437573746f6d5573657244617461',
+    );
+    expect(resultPaymentId?.length).toBe(paymentIdDigits);
+    expect(resultPaymentId).toMatch(paymentIdRegex);
+  });
   it('Simple parse OpReturn 1 no paymentId', () => {
```
BytesOfMan, line 160: `it('Encodes a PayButton v0 protocol transaction with custom user data and no nonce')`

react/src/tests/opReturn.test.ts
```diff
@@ -85,94 +107,109 @@ describe('getDataPushdata', () => {
...
-  })
+    const fullResult = encodeOpReturnProps('myCustomUserData', true); // 16 bytes
+    expect(fullResult).toBe(
+      allResultsPrefix +
+        '10' + // 16 in hex
+        '6d79437573746f6d5573657244617461' +
+        '00', // Empty paymentId pushdata
+    );
+  });
   it('Simple parse OpReturn 2', () => {
```
BytesOfMan, line 169: `it('Encodes a v0 PayButton protocol tx with data of expected keyvalue shape and nonce')`

react/src/tests/opReturn.test.ts
```diff
@@ -85,94 +107,109 @@ describe('getDataPushdata', () => {
...
-    const fullResult = parseOpReturnProps('my=Longer more=sensible|user|data') // 33 bytes
-    const resultPaymentId = fullResult?.slice(-paymentIdDigits)
-    const resultNoPaymentId = fullResult?.slice(0,-paymentIdDigits)
-    expect(resultNoPaymentId).toBe(allResultsPrefix +
-      '21' + // 33 in hex
-      '6d793d4c6f6e676572206d6f72653d73656e7369626c657c757365727c64617461')
-    expect(resultPaymentId?.length).toBe(paymentIdDigits)
-    expect(resultPaymentId).toMatch(paymentIdRegex)
-  })
+    const fullResult = encodeOpReturnProps('my=Longer more=sensible|user|data'); // 33 bytes
```
BytesOfMan, line 170: why is the data like this here? should be described somewhere.

react/src/tests/opReturn.test.ts
```diff
@@ -85,94 +107,109 @@ describe('getDataPushdata', () => {
...
+    const resultNoPaymentId = fullResult?.slice(0, -paymentIdDigits);
+    expect(resultNoPaymentId).toBe(
+      allResultsPrefix +
+        '21' + // 33 in hex
+        '6d793d4c6f6e676572206d6f72653d73656e7369626c657c757365727c64617461',
+    );
+    expect(resultPaymentId?.length).toBe(paymentIdDigits);
+    expect(resultPaymentId).toMatch(paymentIdRegex);
+  });
   it('Simple parse OpReturn 2 no paymentId', () => {
```
BytesOfMan, line 181: `it('Encodes a v0 PayButton protocol tx with data of expected keyvalue shape and no nonce')`

react/src/tests/opReturn.test.ts
BytesOfMan: ok. Ideally we should have a standardized way to encode/decode this data in the spec. utf8 is probably not the best way to do this but it will work.

However, this data can be anything (for now) per spec. It's up to PayButton to optimize how the space is used.

Probably do not want to use space characters (i.e. what is now separating `my=Long` from `more=sensible...`) as the standard delimiter since I can imagine spaces being common in string values. What if the user wants my to actually `my="Longer word than just Longer`? Will this be handled by the current parsing?

Don't need a complicated scheme for encoding and decoding this data but this will also need a unit tested function with all of the cases you support. Outside scope of this diff but just as important to ensuring the txs are useful once they start firing.

react/src/tests/opReturn.test.ts
```diff
@@ -85,94 +107,109 @@ describe('getDataPushdata', () => {
...
+      true,
+    ); // 33 bytes
+    expect(fullResult).toBe(
+      allResultsPrefix +
+        '21' + // 33 in hex
+        '6d793d4c6f6e676572206d6f72653d73656e7369626c657c757365727c64617461' +
+        '00', // Empty paymentId pushdata
+    );
+  });
   it('Throws if too long', () => {
```
BytesOfMan, line 193: `it('Does not support data param exceeding 205 bytes if nonce is present')`

react/src/tests/opReturn.test.ts
```diff
@@ -85,94 +107,109 @@ describe('getDataPushdata', () => {
...
-    expect(() =>
-      parseOpReturnProps(data208bytes)
-    ).toThrow('Maximum 205 byte size exceeded for user data: 208')
-  })
+    const data208bytes = '😂'.repeat(52); // 52 * 4 = 208 bytes
+    expect(() => encodeOpReturnProps(data208bytes)).toThrow(
+      'Maximum 205 byte size exceeded for user data: 208',
+    );
+  });
   it('Is not too long if no paymentId', () => {
```
BytesOfMan, line 199: `it('Supports data push exceeding 205 bytes if no nonce')`

need to add these tests

1. No nonce, max data push
2. No nonce, data push still too big

react/src/util/opReturn.ts
```diff
@@ -82,24 +95,29 @@ function getDataPushdata(data: string, disablePaymentId=false) {
...
 // `08` - pushdata for the optional paymentId (paymentId), signifying this tx has 8 bytes of paymentId data
 // `0102030405060708` - The 8-byte paymentId (paymentId)
+//
+// Result: 0450415900001668656c6c6f20776f726c64080102030405060708
 
-export function parseOpReturnProps(
+export function encodeOpReturnProps(
   opReturn: string | undefined,
-  disablePaymentId=false
+  disablePaymentId = false,
```
BytesOfMan, line 103: this param is super confusing

we have a negative value that is by default negative

So by default, we have a payment id.

Also confusing, is `opReturn` here the data push?

These params should be matched with the spec, `encodePayButtonTx(dataPush=false, nonce=false)`

Calling `encodePayButtonTx()` should give `0450415900000000` and there should be a unit test for this.

BytesOfMan (changes requested):
We have a lot going on in this pull request. Unit tests named numerically makes it difficult to confirm that all cases are covered.

tbh could benefit from a refactor now that we know all the parts. This PR has evolved incrementally, adding a new thing or correcting an old thing, and now there is some spaghetti.

pseudocode for a much cleaner function

```
/**
 * @param {string} data, the data push as text, to be utf8 encoded and added to OP_RETURN per v0 spec
 * @param {boolean} nonce if true, include 8 bytes of random data. If false the nonce push will be '00'
 * @throws {Error} if data push exceeds available byte space in OP_RETURN
 * @returns {string}
encodePayButtonTx(data='00', nonce=false) {
const initialChars = `045041590000` // always true, should be stored as a constant in a file that stores constants, or initialized in page as `const PAYBUTTON_V0_START_CHARS = `045041590000`

if data is not '00', dataFragment = getDataFragment(data), else dataFragment = '00'

if nonce, nonceFragment = getNonceFragment, else nonceFragment = '00'

const op_return_raw = initialChars + dataFragment + nonceFragment

const OP_RETURN_RAW_MAX_BYTES = 222
if op_return_raw.length > OP_RETURN_RAW_MAX_BYTES, throw

return op_return_raw
}

File: react/src/util/opReturn.ts (hunk 19 above)
Comment (Klakurka) [reply react/src/util/opReturn.ts:103]: I'm open to reversing the negatively-named options but if we do that then we need to be consistent about it and update the others which do in another PR.

The reason for the negatively-named options in the first place is because it's probably easier for people who aren't developers.

Comment (chedieck) [reply react/src/util/opReturn.ts:103]: Yes, since having the `paymentId` is the default, an user will probably look into this parameter (disablePaymentId) when he thinks "how can I get rid of that thing", so having a "disablePaymentId" makes it super clear that that is the option they were looking for.

`opReturn`  is indeed the NULL DATA script dataPush but I decided calling it `opReturn` because it was already what was being done in the telegram channels/foruns/twitter I looked into, as a sort of metonymy for 'the datapush that follows an op_return op code'.  Wouldn't be complicated to rename it all (here and in the server) to `dataPush`, but that definetively would have to be done in a separate issue.

That is also why the function name as of rn is `encodeOpReturnProps`, because it encodes the props `op-return` that the user has passed.  I think calling it `encodePayButtonTx` is not precise, since we are not encoding any tx data besides the `opReturn` (the dataPush, in more correct terms); and we also do encode Txs in another way in the server (using the tx data + a private key to generate a hmac that the user could use to confirm that the POST request about the tx really came from us, and is not spoofed).

As for the other comments, regarding the unittests, I'll make sure to fix all the issues you pointed out and also address it so that their names are more clear about their description and that they match one-to-one the different test cases for what their respective functions propose. 

And finally, regarding the space separated values, indeed it currently will not even accept a syntax like `myKey='quoted string with spaces'`; so that would be a problem. We also support JSON, which would cover this case, but in a more byte-costly way.

So I think we should actually discuss the specs for this key=value parsing feature more thoroughly and arrive at something like the OP_RETURN data push specs, before moving forward with it.

Comment (Klakurka) [reply react/src/util/opReturn.ts:103]: Well we talked about supporting spaces with the "simple" PayaButton parsing and decided to just not support it for now, either requiring them to use JSON or supporting it later when there's actually some proven demand for it.

There's really no issue with just doing it later because the on-chain data would be the same. The only difference would be on the server where the submitted OP_RETURN would be updated from just a simple string (where we just give up trying to parse it) to an object that we were able to parse.

File: react/src/tests/opReturn.test.ts (hunk 18 above)
Comment (chedieck) [reply react/src/tests/opReturn.test.ts]: added.

File: react/src/tests/opReturn.test.ts (hunk 10 above)
Comment (chedieck) [reply react/src/tests/opReturn.test.ts]: No, this is the slice of the part that only contains the data. The slice that contains the paymentId contains its pushdata; two lines above. I've changed the variable names to make it more clear.

**Comment by chedieck** [comment]
@BytesOfMan I redid all tests names, addressed the mentioned missing cases to be covered, and renamed some variables  in the  [last commit](https://github.com/PayButton/paybutton/pull/340/commits/aec3b20217bf04b8b6b60b166e21f1a582db001c), let me know what you think

File: react/src/tests/opReturn.test.ts [hunk 20]
```diff
@@ -1,82 +1,122 @@
...
+    const prefixedPaymentId = generatePushdataPrefixedPaymentId(8);
+    expect(prefixedPaymentId.length).toBe(18); // 8 * 2 (two hex chars=byte) + 2(pushdata prefix)
+    expect(prefixedPaymentId.slice(0, 2)).toBe('08');
+  });
+  it('Generates 16 byte paymentId', () => {
+    const prefixedPaymentId = generatePushdataPrefixedPaymentId(16);
+    expect(prefixedPaymentId.length).toBe(34); // 16 * 2 + 2
+    expect(prefixedPaymentId.slice(0, 2)).toBe('10');
+  });
+  it('Generates paymentId at 75 bytes limit', () => {
```
BytesOfMan, line 74: isn't the paymentId always 8 bytes, to spec? 

Does the function support this as variable because this could change in the future?

Could just hardcode it in the function and have only 1 or 2 unit tests

react/src/tests/opReturn.test.ts
```diff
@@ -85,94 +125,102 @@ describe('getDataPushdata', () => {
...
-    ).toThrow('Maximum 214 byte size exceeded for user data: 216')
-  })
-})
+  const allResultsPrefix = '04' + '50415900' + '00';
+  const paymentIdDigits = 18;
+  const paymentIdRegex = /^08[0-9a-fA-F]{16}$/;
+  it('Encodes undefined data with paymentId', () => {
+    const fullResult = encodeOpReturnProps(undefined);
+    const resultData = fullResult?.slice(0, -paymentIdDigits);
+    expect(resultData).toBe(allResultsPrefix + '00'); // 00 is pushdata for no data
```
BytesOfMan, line 141: Would help to have a comment above this line, something like

// Test the string through the data push, as paymentId is expected to be random every time

react/src/util/opReturn.ts
BytesOfMan: > There's really no issue with just doing it later because the on-chain data would be the same.

Ideally PayButton won't have to parse all kinds of different data. Option is available for PayButton to control own destiny here and avoid this issue. 

Would be nice if this function is

1 - The only function creating v0 paybutton tx op_return outputs
2 - It only creates them in one format that can always be parsed for v0

That way if you can still change the format by changing the version  number. but you don't need to parse v0 txs in multiple different ways. Is it JSON? Is it utf8 string? What if there's another idea and it's very tempting to change it without changing the version?

It's all fixable but I could see it causing a lot more work for PayButton down the road.

react/src/util/opReturn.ts
```diff
@@ -82,24 +95,29 @@ function getDataPushdata(data: string, disablePaymentId=false) {
 // `68656c6c6f20776f726c64` - data payload, 'hello world' as ASCII encoded to hex
 // `08` - pushdata for the optional paymentId (paymentId), signifying this tx has 8 bytes of paymentId data
 // `0102030405060708` - The 8-byte paymentId (paymentId)
+//
+// Result: 0450415900001668656c6c6f20776f726c64080102030405060708
 
-export function parseOpReturnProps(
+export function encodeOpReturnProps(
```
BytesOfMan, line 101: Should link to the spec in the documentation for this function (i.e. just a comment above would be good)
`//https://github.com/Bitcoin-ABC/bitcoin-abc/blob/master/doc/standards/paybutton.md`

BytesOfMan (approved):
feedback at this point is nits. 

I do think it's important to settle on a solid format for the data. It is possible to parse anything, but you could also just parse one thing, which is much easier. You could still support other types going forward by changing the version from `00` to, say, `0101`

Then easy for the indexer running on your server to identify different paybutton txs and know what kind of parsing they need.

react/src/util/opReturn.ts
BytesOfMan: > That is also why the function name as of rn is encodeOpReturnProps, because it encodes the props op-return that the user has passed. I think calling it encodePayButtonTx is not precise, 

well -- the spec is for v0 paybutton txs, which is the only version this function handles. At least document it in the function comments.

If there is ambiguity into what is or is not a v0 paybutton tx, should be addressed in the spec. we're assuming that the tx is a valid ecash tx -- what makes it a paybutton v0 tx is the op_return output conforming to the documented spec

Klakurka: > Would be nice if this function is the only function creating v0 paybutton tx op_return outputs

It should be - I'm not sure if things will get more complicated if we're eventually doing BIP 70 but I'm pretty sure we will still be able to keep it just within the button.

> It only creates them in one format that can always be parsed for v0

1. We do want to at least allow entering simple bits of text as simply entering JSON isn't practical for much of the target audience (who are often not developers).
2. The spec for the "data" portion is just "here's a bucket of data". If we can successfully parse it into JSON on the server, that's good but if not it's not a big deal. We can't guarantee good data in the OP_RETURN anyway since it's permissionless -- also typos are a thing.
3. Every app using PayButton is going to be different. Some people will have weird (to us) systems for managing things. We can encourage people to do stuff a certain way but limiting them when they really really want to do it some other way is probably a bad idea in this case. I'm reminded of CoinMarketCap and how they explicitly requested that the API we built for them return "just the number" because they couldn't handle parsing JSON (which is what we provided them with originally). So in our case here, they'll either be getting back a JSON object or a simple string in their payment trigger (server webhook) if they have that set up -- and they can do whatever they want with it at that point.

If a 3rd party service wanted to go through and index paybutton txs, they should be able to follow the spec, just treating the "data" portion as a big string.

Klakurka: The idea of trying to parse 2 things is mainly to give people options who find JSON scary -- which might sound like a dumb reason but it's a real problem for people who aren't developers but who will want to do things like sorting, filtering, etc of transactions through the PayButton (server) UI.

I think it could be worth thinking about doing the JSON parsing attempt on the client button instead of having the server do it, but there's the tradeoff of JSON taking up more space. Still worth discussing.

BytesOfMan: > We do want to at least allow entering simple bits of text as simply entering JSON isn't practical for much of the target audience (who are often not developers).

imo it's risky having this -- why not condition all the input data to meet your spec? if json is intimidating, can have multiple 'key' and 'value' fields, for example. Even without this, could still have some sort of preprocessing between what the user is entering and what is encoded in the tx.

Or, even simpler things, like "strings cannot have spaces", or "replace all spaces in strings with hex char for space"

chedieck: We could use a query parser such as  `qs` and `JSON.parse`. So we just try both and return whatever works, or a string. I don't think they ever overlap, and if they do, will be probably in edge cases in which we will just return an empty string anyway.

This also would be easy to document, since it already is.

Klakurka: Actually sounds like a decent idea to me, rather than rolling our own and attempting to fix it up later.

@BytesOfMan thoughts?

BytesOfMan: Yeah sounds like a good way to validate the user input. Only accept user input that passes the parser you want to use.

it's not really a spec thing, but you want some kind of limitation on what type/shape of data gets into the tx.

react/src/util/opReturn.ts
chedieck: I've redone the tests in https://github.com/PayButton/paybutton/pull/353; since I had to change the implementation. I added this comment there.

Klakurka (approved):
We're going to do more using the query string lib on the client but we'll do that in another branch.

PR #341, opened 2024-01-19
[#302] feat: disable payment id props

chedieck:
Related to #302

Depends on
---
- [ ] #340

Description
---
Implements `disablePaymentId` props & tests for it.

Test plan
---
Make sure the generated buttons when this option is on/off doesn't include/includes the nonce at the end.


Klakurka (changes requested):
Getting this on build: 
image

chedieck:
Should be fixed now.

Klakurka:
The `08` and `0c` pushdata bytes are at the end of each of their sections but the `04` one comes before -- why?

chedieck:
> The `08` and `0c` pushdata bytes are at the end of each of their sections but the `04` one comes before -- why?

Where is that ?

Klakurka:
Here's an example PayButton hex output:

`04504159000025636c6173734f663d323031332062756c6c59656172733d323031337c323031377c323032310864e6de49eda7d667`

Actually I don't see the `0c` on this one at all (shouldn't it be to the left of the payment ID?

I see the `08` but it's on the right of the data.

chedieck:
`04504159000025636c6173734f663d323031332062756c6c59656172733d323031337c323031377c323032310864e6de49eda7d667`

1. The first 04 is pushdata for the protoocol, meaning the next 4 bytes are the protocol (50415900)
2. Then there is a 00 which is the version byte
3. Then there is `25` meaning the next `2*16 + 5 = 37` bytes are the data: (636c6173734f663d323031332062756c6c59656172733d323031337c323031377c32303231)
4. Then there is 08 meaning the next 8 bytes are the payment Id `64e6de49eda7d667`

> Actually I don't see the `0c` on this one at all (shouldn't it be to the left of the payment ID?

No, the paymentId follows `08`, because its 8 bytes long. I see there is `0c` in the specs, which would be the pushdata for a 12 byte paymentId, so we probably should put 08 there in the example.

Klakurka (dismissed):
The server still saves `paymentID: ""` which we can probably omit but that's the server's responsibility.

Just make sure we don't lose any of the docs updates we made which would come in from the other branch.

PR #342, opened 2024-01-19
[#302] Refactor: rename nonce -> paymentId

chedieck:
Related to #302

Depends on
---
- [x] #341
- [x] #345

Description
---
Simple refactor, renames `nonce` to `paymentId` everywhere to avoid having two names for the same things since we use `paymentId` in the props and in the server.

Test plan
---
Nothing should change.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -666,7 +666,7 @@ opReturn = "myCustomMessage"
 
 ## disable-payment-id
 
-> **The ‘disable-payment-id’ removes the random id generated for the payment that will be send with the transaction
+> **The ‘disable-payment-id’ parameter removes the random ID generated for the payment that is used to prevent the onSuccess callback to be triggered by a person who has the payment screen open at the same time as another.**
 
 ?> This parameter is optional. Default value is false. Possible values are true or false.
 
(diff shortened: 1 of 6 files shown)


Klakurka (changes requested):
I changed the payment ID option description in English. Can you re-do the Chinese translations for those? Try to keep the formatting the same as the other examples.

Also I'd prefer if the 'disable' options are all together so maybe worth moving it down.

chedieck:
> I changed the payment ID option description in English. Can you re-do the Chinese translations for those? Try to keep the formatting the same as the other examples.
> 
> Also I'd prefer if the 'disable' options are all together so maybe worth moving it down.

This would probably best have been done in [the other PR](https://github.com/PayButton/paybutton/pull/341); nonetheless all addressed.

Klakurka (approved):
Yep -- I realized after moving on from this one.

Klakurka (dismissed):

chedieck:
To facilitate review I created https://github.com/PayButton/paybutton/pull/345 that should be reviewed and merged first to make the diff here only account for the string substitution related to the PR.

Klakurka (approved):

Issue #331, opened 2024-01-14
QR code no longer shows up when there's a failure on the server (affects BCH)


Klakurka:
Even though we can't detect payments, we shouldn't actually stop them from paying.

This used to work but for some reason isn't anymore.

<img width="305" alt="image" src="https://github.com/PayButton/paybutton/assets/4713204/8c194859-e708-4935-adf5-724411aa77a5">

PR #344, opened 2024-01-20
fix: catch api error

lissavxo:
Related to #331

Description
---
This issue was happening because the app did not set the load to false when the api call fails, so in this fix we 
add error handling and set the loader to false to fix qrcode not showing up 

Test plan
---

- Run story book using this steps from the [docs](https://docs.paybutton.org/#/?id=developer-quick-start:~:text=of%20the%20project.-,Developer%20Quick%20Start,-Build)

```
> docker-compose up
> Open http://localhost:6006.
```

- Open storybook 
- Go to `Paybutton` 
- Set `to` option to a valid BCH address like this one  `bitcoincash:qrmm7edwuj4jf7tnvygjyztyy0a0qxvl7q9ayphulp` 
- Open the Payment dialog by clicking in the `Paybutton`  and check if the `BCH` `Qrcode`  is showing up as expected

### DIFF
--- a/react/src/components/Widget/Widget.tsx
+++ b/react/src/components/Widget/Widget.tsx
@@ -213,6 +213,7 @@ export const Widget: React.FC<WidgetProps> = props => {
     (async (): Promise<void> => {
       const balance = await getAddressBalance(to, apiBaseUrl);
       setTotalReceived(balance);
+      setLoading(false);
     })();
   }, [newTxs]);
 
--- a/react/src/util/api-client.ts
+++ b/react/src/util/api-client.ts
@@ -31,9 +31,15 @@ export const setListener = (socket: Socket, setNewTxs: Function): void => {
 export const getAddressBalance = async (
   address: string,
   rootUrl = config.apiBaseUrl,
-): Promise<number> => {
-  const res = await axios.get(`${rootUrl}/address/balance/${address}`);
-  return isNaN(res.data) ? null : res.data;
+): Promise<number | undefined> => {
+  try {
+    const res = await axios.get(`${rootUrl}/address/balance/${address}`);
+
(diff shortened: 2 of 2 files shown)


Klakurka (approved):
Looks good!

react/src/components/Widget/Widget.tsx
```diff
@@ -213,6 +213,7 @@ export const Widget: React.FC<WidgetProps> = props => {
     (async (): Promise<void> => {
       const balance = await getAddressBalance(to, apiBaseUrl);
       setTotalReceived(balance);
+      setLoading(false);
```
chedieck, line 216: Can you justify this? I see that there is an effect below that will set this to false when `totalReceived` changes (which is changing here), so I am afraid that this can lead to loading being set to false too early.

chedieck (changes requested):

react/src/components/Widget/Widget.tsx
lissavxo: do we need the `AddressBalance` to show the `Qrcode` up and stop the loading? 
because I see in the PaymentDialog that the effect is not displayed and it shows the qrcode even  this function getAddressBalance returning undefined for bitcoin addressess 

you can test it using this branch that has a stable build,  

- run the storybook (docker-compose up) 
- go to payment dialog set a bitcoincash address, it will show the qrcode 

if you remove this extra line I added ` setLoading(false); ` and go to the `paybutton` component and set a bitcoincash address (e. g. bitcoincash:qrmm7edwuj4jf7tnvygjyztyy0a0qxvl7q9ayphulp) and hit the button, you will see that it loads infinitely because the effect is not working in the payment dialog

chedieck (approved):

PR #353, opened 2024-01-25
[#302] feat: payment id on components

chedieck:
Related to #302

Depends on
---
- [x] #340 

Description
---
Adds the `paymentId` to the `PayButton` component to be passed as props to the other components; and as arguments to the `onOpen`  and `onClose` functions.

If no props arrive (as  of when creating just the Widget), the paymentId will be generated randomly in the Widget itself.

Test plan
---
Try to see the paymentId on the `onOpen` and `onClose` functions.

### DIFF
--- a/paybutton/src/index.tsx
+++ b/paybutton/src/index.tsx
@@ -110,9 +110,9 @@ function renderDialogButton(dialogbuttonExists: boolean): void {
 function openDialog(props: PaymentDialogProps): void {
   const container = document.createElement('div');
   document.body.appendChild(container);
-  const onClose = (paymentId?: string, success?: boolean) => {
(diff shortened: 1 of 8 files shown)


Klakurka (changes requested):
I think payment ID should go at the end rather than the beginning of the props since it won't be as commonly used and it doesn't always exist.

Klakurka (changes requested):
The value loads in right, but then it never changes when you close the dialog and re-open it.

Pretty sure it should be updating every time right?

I'm not sure if we're covering this case but if a payment is made to a widget, the payment ID should change after it's done if they decide to update the amount or something -- can maybe have this be its own task.

Klakurka (approved):
https://github.com/PayButton/paybutton/issues/358

Issue #357, opened 2024-01-26
onOpen & onClose function args & docs


chedieck:
The args they take should be in the docs.

Also, gotta make sure it is clear if e.g. the type of function `onOpen` it takes is `(arg1?: string, arg2?: string) => ...` or `(arg1:string, arg2: string) => ...` or both; not sure yet

chedieck:
Maybe make `to` and `paymentId`mandatory since they are always well defined (paymentId is defined even if it's disabled; disabling it changes ANOTHER variable) and `amount` optional since it may not exist, but I haven't gave it much thinking yet

PR #360, opened 2024-01-27
fix(357): fix documentation

lissavxo:
Related to #357

Description
---
- added parameters to the documentation

Test plan
---
run `yarn start:docs` and check if the documentation on `onOpen` and `onClose` has the paremeters

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -581,7 +581,11 @@ onTransaction = transactionCallback
 
 > **The ‘on-open’ parameter specifies the callback function that runs when the button is clicked, before the dialog opens.**
 
-?> This parameter is optional. Default value is empty. Possible values are any defined function.
+#### *callback* arguments
+
+- amount: *number* - how much money is being requested
+- paymentId: *string* - the unique identifier for the payment
+- to: *string* - where the money will be sent to
 
 **Example:**
 <!-- tabs:start -->
@@ -611,6 +615,11 @@ onOpen = successCallback
 > **The ‘on-close’ parameter specifies the callback function that runs when the dialog is closed.**
 
 ?> This parameter is optional. Default value is empty. Possible values are any defined function.
+
+#### *callback* arguments
+
+- **success**: *boolean* - status of the payment
+- **paymentId**: *string* - the paymentId
 
 **Example:**
 <!-- tabs:start -->
--- a/docs/zh-cn/README.md
+++ b/docs/zh-cn/README.md
@@ -581,6 +581,12 @@ onTransaction = transactionCallback
 
 ?> 这个参数是可选的。默认值为空。可能的值是任何已定义的函数。
 
+#### *回调* 参数
+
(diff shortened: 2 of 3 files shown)


docs/README.md
```diff
@@ -577,11 +577,16 @@ onTransaction = transactionCallback
 
 <!-- tabs:end -->
 
-## on-open
+## on-open ( *optional* )
```
Klakurka, line 580: Let's remove the `( optional )` bit here as it's unclear _what_ is optional to somebody reading this for the first time.

docs/README.md
```diff
@@ -606,12 +611,17 @@ onOpen = successCallback
 
 <!-- tabs:end -->
 
-## on-close
+## on-close  ( *optional* )
```
Klakurka, line 614: Let's remove the `( optional )` bit here as it's unclear _what_ is optional to somebody reading this for the first time.

docs/README.md
```diff
@@ -577,11 +577,16 @@ onTransaction = transactionCallback
...
 
 > **The ‘on-open’ parameter specifies the callback function that runs when the button is clicked, before the dialog opens.**
 
-?> This parameter is optional. Default value is empty. Possible values are any defined function.
+#### *callback* arguments
+
+- **expectedAmount**: *number* - returns the amount set for the paybutton
+- **address**: *string* - returns address if set in the paybutton element
+- **paymentId**: *string* - returns paymentId if set in the paybutton element
+- **to**: *string* - returns destination address if set in the paybutton element
```
Klakurka, line 589: Let's remove the `for/in the paybutton` bit on each of these as it's a bit redundant given that these are the PayButton docs.

Klakurka (changes requested):
- As with the other branch, we'd categorize this under feat not fix as it's not a bug -- no need to worry about this though, just something for next time.
- The Chinese translations will need to be done. I can do them if you're not confident in getting them right.

docs/README.md
```diff
@@ -581,7 +581,12 @@ onTransaction = transactionCallback
 
 > **The ‘on-open’ parameter specifies the callback function that runs when the button is clicked, before the dialog opens.**
 
-?> This parameter is optional. Default value is empty. Possible values are any defined function.
+#### *callback* arguments
+
+- **expectedAmount**: *number* - the amount set
+- **address**: *string* - your address
+- **paymentId**: *string* - the paymentId
+- **to**: *string* - the destination address
```
Klakurka, line 589: I think we might have made some mistakes when implementing this change originally.

These should be:

`amount` (not `expectedAmount`)
`paymentId`
`to` (no `address`)

So my suggestion is to change it to these ^ for now and we'll make a new PR for fixing these up to be in line with the rest of the site.

Klakurka (changes requested):
We can fix the actual argument naming here: #363

Klakurka (approved):

Issue #359, opened 2024-01-27
Deprecate (set to false by default) randomSatoshis


Klakurka:

Klakurka:
Actually this seems to be not working right... can you have another look?

PR #361, opened 2024-01-27
fix(359): randomSatoshis false

lissavxo:
Related to #359

Description
---

set randomSatoshis to false


Klakurka (changes requested):
Update the docs as well.

Klakurka (changes requested):
- We make sure to have the example code for each option be different from the default so it can be copy + pasted and have an actual effect, so let's change the example here to true instead of false.
- Can you go in and update the Chinese docs as well? In this case it's just simply a matter of swapping true <-> false.

Klakurka (approved):

Issue #369, opened 2024-01-30
Widget not firing onSuccess function on payment detection


Klakurka:
The toast works, but it doesn't run the onSuccess function.

PR #371, opened 2024-01-30
fix(#369): call on-succes

lissavxo:
Related to #369 

Description
---
fix call on-succes conditions
 fixed amount not being set correctly after changign

Test plan
---

Open the project run `yarn build` , do a transaction with the amount indicated by the button and it should trigger the onSuccess function you set

Test 1: 
 set a button passing the `on-success` function > do the payment > check if the onSuccess function was triggered

Test 2: 
 set a widget passing the `on-success` function >  do the payment >  check if the onSuccess function was triggered 
 
 Test 3
  set a widget passing the `on-success` function   and set an `amount` > change the amount manually> proceed with the payment >   check if the onSuccess function was triggered

### DIFF
--- a/react/src/components/Widget/WidgetContainer.tsx
+++ b/react/src/components/Widget/WidgetContainer.tsx
@@ -161,7 +161,7 @@ export const WidgetContainer: React.FunctionComponent<WidgetContainerProps> =
         if (
           cryptoAmount &&
           receivedAmount.isEqualTo(new BigNumber(cryptoAmount)) &&
-          txPaymentId === paymentId
+          (paymentId ? txPaymentId === paymentId : true)
         ) {
           setSuccess(true);
           onSuccess?.(transaction.id, receivedAmount);
--- a/react/src/components/PaymentDialog/PaymentDialog.tsx
+++ b/react/src/components/PaymentDialog/PaymentDialog.tsx
@@ -73,10 +73,6 @@ export const PaymentDialog = (
     if (onClose) onClose(success, paymentId);
     setSuccess(false);
   };
-  const handleSuccess = (txid: string, amount: BigNumber): void => {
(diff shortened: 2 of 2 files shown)


react/src/components/Widget/WidgetContainer.tsx
```diff
@@ -218,17 +219,17 @@ export const WidgetContainer: React.FunctionComponent<WidgetContainerProps> =
...
 
     return (
       <React.Fragment>
         <Widget
           to={to}
           {...widgetProps}
-          amount={amount}
-          setAmount={setAmount}
+          amount={cryptoAmount}
+          setAmount={setCryptoAmount}
```
Klakurka, line 232: What did renaming these do for us?

Klakurka (changes requested):
- NAN showing on payment dialog now when there's a fiat amount
image
-

react/src/components/Widget/WidgetContainer.tsx
lissavxo: we already had these values in the component here: 
`const [cryptoAmount, setCryptoAmount] = useState<string>();` 

I think they mean the same thing as amount and setAmount, just were named differently to avoid errors with props with the same name ? 

because of the useEffects not correctly set the component wasn't passing the values correctly so I did this fix to address correctly the value of the amount

Klakurka (approved):
- The 'close' button seems to be missing on the payment dialog once the payment has been made: 
image #375
- onSuccess provides something called 'txid' as well as an 'amount' but this isn't done the way I was expecting. If we could display the actual blockchain txid, that would be totally fine, but the amount should just be a simple number (amount in XEC). 
image #374

As far as the scope of this PR goes though, looks good.

Issue #375, opened 2024-01-31
'Close' button missing from payment dialog after successful payment


Klakurka:
<img width="273" alt="image" src="https://github.com/PayButton/paybutton/assets/4713204/b63e1792-1994-4e65-9b0a-cef2b2d449d1">

Issue #374, opened 2024-01-31
Update on-success function data to make more sense


Klakurka:
onSuccess provides something called 'txid' as well as an 'amount' but this isn't done the way I was expecting. If we could display the actual blockchain txid, that would be totally fine, but the amount should just be a simple number (amount in XEC). 

<img width="323" alt="image" src="https://github.com/PayButton/paybutton/assets/4713204/32f30cde-f994-4ed1-87bb-9c03e5308cbe">

PR #376, opened 2024-02-01
onSuccess fixes

lissavxo:
Related to #375  #374 

 

Description
---
This PR will introduce these fixes: 
- fix set success to true in paymentDialog
- fix calling onSuccess when no amount set 
- Set amount to number in onSuccess callback 

Test plan
---
Open the project, run `yarn build`  and test if onSuccess function is called and if it returns the amount properly

@lissavxo

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -525,6 +525,12 @@ successText = "Thanks!"
 
 ?> This parameter is optional. Default value is empty. Possible values are any defined function.
 
+#### *callback* arguments
+
+- hash: *string* - the hash of the broadcasted transaction
+- amount: *string* - how amount money is being requested
+- paymentId: *string* - the payment unique identifier
+
 **Example:**
 <!-- tabs:start -->
 
@@ -553,6 +559,13 @@ onSuccess = successCallback
 > **The ‘on-transaction’ parameter specifies the callback function that runs upon any payment received to the specified address.**
 
 ?> This parameter is optional. Default value is empty. Possible values are any defined function.
+
+#### *callback* arguments
+
+- hash: *string* - the hash of the broadcasted transaction
+- amount: *string* - how amount money is being requested
+- paymentId: *string* - the payment unique identifier
+
 
 **Example:**
 <!-- tabs:start -->
--- a/docs/README.md
+++ b/docs/README.md
@@ -525,11 +525,12 @@ successText = "Thanks!"
 
 ?> This parameter is optional. Default value is empty. Possible values are any defined function.
 
-#### *callback* arguments
+#### *callback*  arguments
 
-- hash: *string* - the hash of the broadcasted transaction
-- amount: *string* - how amount money is being requested
-- paymentId: *string* - the payment unique identifier
+- **transaction** (`object`): 
+  - **hash** (`string`): The hash of the broadcasted transaction
+  - **amount** (`string`): Amount of money requested
+  - **paymentId**\* (`string`): Unique identifier for the payment
 
 **Example:**
 <!-- tabs:start -->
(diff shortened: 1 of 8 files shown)


Klakurka:
Why is removing the big number lib necessary here? What was wrong with it?

Klakurka (changes requested):
I think we talked about this before but either I've changed my mine or didn't really think about it clearly in the first place but until we have a dedicated non-success "transaction" toast (and success chime sound), let's block the toast if the payment ID doesn't match.

lissavxo:
> Why is removing the big number lib necessary here? What was wrong with it?

We removed BigNumber from form the Amount param that was being sent in Onsucces function. Now we are returning a number instead. 
 
BigNumber  returns a object so we changed it to be a number instead to have only the amount sent.

we can also adjust to try to extract the number from the BigNumber object, so it matches the value we are showing in the toast.

Klakurka (changes requested):
- The 'txid' that's provided in the onSuccess function is... I'm not really sure what it is exactly. It should be the actual broadcasted txid.
- It would be nice to have the Payment ID be available too alongside txid and amount.
- Let's have the success toast only fire using the same logic as the check mark.

docs/README.md
```diff
@@ -525,6 +525,12 @@ successText = "Thanks!"
 
 ?> This parameter is optional. Default value is empty. Possible values are any defined function.
 
+#### *callback* arguments
+
+- hash: *string* - the hash of the broadcasted transaction
+- amount: *string* - how amount money is being requested
+- paymentId: *string* - the payment unique identifier
+
```
Klakurka, line 533: After actually looking at the data payload coming from the server and thinking about this for a bit, I've had a change of opinion on this.

Why not just have only the tx object be available here and they can dig into it and use whatever they need from that?

This way, we wouldn't have to go back in and add/remove/update the arguments when they change, right?

Klakurka (changes requested):

docs/README.md
Klakurka: I made this for the server as we're sending a lot of useless data to the client: https://github.com/PayButton/paybutton-server/issues/811 -- maybe we can think about updating the docs to include all of the tx data once that's eventually done? I'm not really concerned about having all of this in the docs beforehand since we made decide to keep/remove other stuff in the payload that I'm not thinking about.

Klakurka: An argument against my idea here of just passing `tx` for them to use is that this is really only applicable to `on-success` and `on-transaction` but not `on-open` and `on-close`.

Thoughts?

Klakurka: Estevao is good with this change so let's go ahead with it in this branch and update the docs again.

react/src/components/Widget/WidgetContainer.tsx
```diff
@@ -144,10 +144,14 @@ export const WidgetContainer: React.FunctionComponent<WidgetContainerProps> =
...
+          amount: transactionAmount, 
+          opReturn, 
+          hash, 
+          paymentId: transactionPaymentId } = transaction;
+        const receivedAmount = new BigNumber(transactionAmount);
 
         const currencyTicker = getCurrencyTypeFromAddress(to);
-
         if (!hideToasts)
           // TODO: This assumes only bch
```
Klakurka, line 156: Is this an old comment or does it still have relevance?

docs/README.md
```diff
@@ -525,6 +525,13 @@ successText = "Thanks!"
 
 ?> This parameter is optional. Default value is empty. Possible values are any defined function.
 
+#### *callback*  arguments
+
+- **transaction** (`object`): 
+  - **hash** (`string`): The hash of the broadcasted transaction
+  - **amount** (`string`): Amount of money being requested
```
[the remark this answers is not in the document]
Klakurka: I also wasn't expecting us to keep the tx 'details' in the docs here but I'm fine with keeping it.

Klakurka (changes requested):

react/src/components/Widget/WidgetContainer.tsx
lissavxo: I believe this was added just for testing

Klakurka: Let's remove it while we're in here.

Klakurka (changes requested):
I can't get "payment success" to happen.

Example payment URI: `ecash:qp2v7kemclu7mv5y3h9qprwp0mrevkqt9gprvmm7yl?amount=10&op_return_raw=045041590000000866d82352aac4c60b`

Klakurka (approved):
OK looks like the server was having issues.

We're still getting toasts appearing when the payment ID doesn't match but we can deal with this in another task.

Issue #380, opened 2024-02-02
Add rollup command


lissavxo:
add command rollup to the root package.json

Klakurka:
I'm pretty confident that this was working before so marking it as a bug.

PR #381, opened 2024-02-02
feat: add start rollup command

lissavxo:
Related to #380 

Description
---
Added the command `start:rollup` to the root of the monorepo

Test plan
---
Open the project, run `start:rollup`


Klakurka (approved):
Actually I don't think I had tested rollup -- good that we have it working now but I've made a new task to update the actual page to include XEC buttons and other new features lime OP_RETURN.

#382

PR #388, opened 2024-02-23
fix: expect no paymentId on widget container

chedieck:
Description
---
Widgets' paymentId was not being correctly used to check for `onSuccess`.

Test plan
---
Make sure that making a tx to a Widget or PayButton that doesn't disable paymentId executes correctly the `onSuccess` function after a payment is made using the copied QR code.


lissavxo (approved):
looks good

PR #392, opened 2024-02-27
[#391] feat: on open uses crypto amount

chedieck:
Related to #391

Description
---
In order to have `onOpen` use the crypto amount, we have to know the crypto price before opening the button. This is now the case, as we fetch the price just as the component loads.

Test plan
---

- Make sure onOpen works with the crypto amount even if the paybutton is fiat,
- Make sure fiat widgets still work normally; i.e they fetch the price too.

### DIFF
--- a/docs/zh-tw/README.md
+++ b/docs/zh-tw/README.md
@@ -594,7 +594,7 @@ onTransaction = transactionCallback
 
 #### *回撥* 參數
 
-amount: number - 請求的金額是多少（以加密貨幣計)
+- amount: number - 請求的金額是多少（以加密貨幣計)
 - **to**: *string* - 付款的唯一标识符
 - **paymentId**: *string* - 资金将汇入的地方
 
--- a/docs/zh-tw/README.md
+++ b/docs/zh-tw/README.md
@@ -594,7 +594,7 @@ onTransaction = transactionCallback
 
 #### *回撥* 參數
 
-- **amount**: *number* - 被請求的金額是多少
+amount: number - 請求的金額是多少（以加密貨幣計)
 - **to**: *string* - 付款的唯一标识符
 - **paymentId**: *string* - 资金将汇入的地方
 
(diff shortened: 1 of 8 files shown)


Klakurka (changes requested):
Seems to work for me.

Nits.

docs/zh-tw/README.md
```diff
@@ -594,7 +594,7 @@ onTransaction = transactionCallback
 
 #### *回撥* 參數
 
-- **amount**: *number* - 被請求的金額是多少
+amount: number - 請求的金額是多少（以加密貨幣計)
```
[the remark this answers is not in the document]
chedieck: Fixed on both

Klakurka (changes requested):
Further, looks like on-open only fires if fiat currency is set. If it's just using a simple XEC amount then it doesn't for some reason.

Klakurka (changes requested):
No longer builds for me:

image

react/src/components/PayButton/PayButton.tsx
```diff
@@ -67,11 +72,33 @@ export const PayButton = (props: PayButtonProps): React.ReactElement => {
...
+  useEffect(() => {
+    priceRef.current = price;
+  }, [price]);
+
+  useEffect(() => {
+    cryptoAmountRef.current = cryptoAmount;
+  }, [cryptoAmount]);
+
+  const waitPrice = async (callback: Function) => {
+    while (true) {
```
lissavxo, line 86: this` while (true) ` loop with `await delay(300) ` could potentially lead to performance issues as it can lock the events loop, would be a better thing to use js tools like setInterval  https://stackoverflow.com/questions/14840527/whiletrue-vs-setintervalfunction-0#:~:text=There%20is%20a%20large%20difference,and%20depends%20on%20the%20browser.
`

chedieck: Refactored

chedieck:
> No longer builds for me:
> image

Try again, should work now

lissavxo:
looks good to me

Klakurka (approved):

lissavxo (approved):

PR #389, opened 2024-02-23
update react 17 to 18

lissavxo:
Related to #

Description
---

Test plan
---

PR #395, opened 2024-03-11
chore: move components to lib folder

lissavxo:
Related to #389 

Description
---
This PR introduces a structure change in the project, we are moving the components from `src` to `lib` so it is more organized and it reflects more that the components are a library, in the future `src` folder can be used to create a demo page for development purposes. 

Test plan
---

open the project, run "yarn clean:build" check if it builds correctly


react/tsconfig.json
```diff
@@ -9,7 +9,7 @@
     "moduleResolution": "node",
     "jsx": "react",
     "sourceMap": true,
-    "declaration": true,
+    "declaration": false,
```
chedieck, line 12: Can you explain me better this decision?

chedieck (changes requested):
Is the purpose of this PR to lay the groundwork for eventually having a single repo logic, instead of the current birepo logic? If so, would then the current `react/` folder become the main repo and the current `paybutton/` folder go to `react/src/`?

Trying to understand better the goal of the PR.

lissavxo:
> Is the purpose of this PR to lay the groundwork for eventually having a single repo logic, instead of the current birepo logic? If so, would then the current `react/` folder become the main repo and the current `paybutton/` folder go to `react/src/`?
> 
> Trying to understand better the goal of the PR.

Well, actually in the PR with the update of the libs we are working only on changing the technologies we are using to more recent ones, we are not making this a single ripo (at least not yet in the PR in progress), the `react` and `paybutton` folders remain independent in the new update.

lissavxo:
> Is the purpose of this PR to lay the groundwork for eventually having a single repo logic, instead of the current birepo logic? If so, would then the current `react/` folder become the main repo and the current `paybutton/` folder go to `react/src/`?
> 
> Trying to understand better the goal of the PR.

the structure is changing inside `react` folder, because it is an independent module so we are refactoring inside of it instead of moving it outside and restructure the monorepo

chedieck:
> Well, actually in the PR with the update of the libs we are working only on changing the technologies we are using to more recent ones, we are not making this a single ripo (at least not yet in the PR in progress), the `react` and `paybutton` folders remain independent in the new update.

I see, just trying to understand better why the refactor.

Klakurka:
I'm not sure if this would be impacted but we need to still be able to build + publish both react and vanilla js versions to npmjs.

react/tsconfig-build.json
chedieck, line 1: Could you brief me on this logic? I mean the creation of this particular file that AFAIK will only be used by `yarn build`

chedieck (changes requested):
Left a question

Klakurka:
Also conflicts.

react/tsconfig-build.json
lissavxo: the difference is that we only want `lib` in the build we do not need to include `src` as it will contain only dev material to test the components while developing, in the extra json config file we include only lib and it will be used only in the build

chedieck (approved):

Issue #378, opened 2024-02-02
Ensure proper thing happens when payment is received


Klakurka:
There's a lot going on right now when a payment is received. I'll try to describe each case here:

Note that both the normal payment dialog and widget should behave almost the same.

When a payment is detected...

onSuccess (including QR code check mark and toast) should fire if:

- Address is correct
- Amount is correct
- OP_RETURN (including Payment ID) is correct

onTransaction (only the function -- nothing else) should fire if:

- Address is correct
- Amount is **NOT** correct
- OP_RETURN is **NOT** correct

What does it mean for an address to be "correct"?

- As we currently only support standard BCH and XEC addresses, it just needs to match. I think we're using a library for this but if they exclude the address prefix then that's still fine.
- When we add support for things like xpub keys and aliases, this will get more complicated.

What does it mean for an amount to be "correct"?

- If an amount is set, it matches the amount received.
- If an amount is not set, then any amount works.

What does it mean for an OP_RETURN to be "correct"?

- BCH doesn't have the same level of standardization of wallets as XEC does so we should simply ignore OP_RETURN there. So when using BCH, all OP_RETURN values are fine. #379
- On XEC, we can simply see if the OP_RETURN matches like we do now. This should still be the case even if `disable-payment-id=true`. It's up to us if we want to have it accept any or even no OP_RETURN if `disable-payment-id=true` and `op-return` is not populated.

PR #403, opened 2024-04-19
Feat/improve payment detected behavior

lissavxo:
Related to #378 

Description
---

- Fixed logic to trigger success and transaction callback
- Fixed sound to play only if the tx was successful
- Isolate logic to trigger success
- Added tests to shouldTriggerSuccess function
- Removed unnecessary comments 

Test plan
---
Open the project, run `yarn clean:build` make a payment and check if we trigger success correctly as described here #378


Klakurka:
> when we have [sound:true hide-toasts: true] will we display sound ?

Yes

lissavxo:
> > when we have [sound:true hide-toasts: true] will we display sound ?
> 
> Yes

Fixed

Klakurka (changes requested):
After changing the amount within an editable dialog, paying to it doesn't trigger on-success.

image

Klakurka (changes requested):
Let's add some tests for this branch.

We can talk about it together with @chedieck to see what we can + should do. It's tedious and unreliable to be testing things manually. The bug with the editable dialog is an example of something that is easy to miss during manual testing.

react/lib/util/api-client.ts
```diff
@@ -1,9 +1,41 @@
...
+  transaction: Transaction,
+  thisPaymentId?: string,
+  cryptoAmount?: string,
+  thisOpReturn?: string,
+) => {
+  const { amount, paymentId, message } = transaction;
+
+  const formattedTxAmount = resolveNumber(amount);
+  const formattedAmountSet = cryptoAmount ? resolveNumber(cryptoAmount) : false;
+  const formattedTxOpReturn = JSON.stringify(message);
```
chedieck, line 26: why are we JSON.stringifying these strings?

chedieck (changes requested):
Left some remarks about test naming, also pushed a commit improving a little on a test

I found that in the button generator:
- Editable toggle set to false still allows it to be editable
- Sending a TX to the address with the same amount, same payment id (in the opReturn) but different data (so different opReturn but same paymentId) triggers onSuccess, not onTransaction. For what I gathered from #378 this is not desired behavior.
- If we have paymentId disabled, then sending the tx with the correct amount and no paymentId still doesn't trigger onSuccess. My suspicion is that the "shouldTriggerOnSuccess" function is comparing undefined with `''` or smth like that.


When solving the bugs above, make sure you first write tests that catch them, then solve them.

Finally, would be good to also:
- add the `yarn test:coverage` command to the root folder
- put min=0 in the amount & goal amount fields of the button generator

react/lib/util/api-client.ts
lissavxo: for some reason sometimes the string is interpreted as an object so I decided to stringfy both to compare. 
we do it with both value because if we compare empty strings without stringfy it will compare this "''" with ''  so we need to do it for both variables

chedieck (changes requested):
onSuccess triggered twice.
изображение

To reproduce the screenshot, go to chedieck.com/paybutton, and send 30XEC to the button with name "ONOPEN".

lissavxo:
> onSuccess triggered twice. изображение
> 
> To reproduce the screenshot, go to chedieck.com/paybutton, and send 30XEC to the button with name "ONOPEN".

this happens on master too, I believe It is due to having a widget and a button with the same address in the same page, they are not isolated. For example, when you pay in the widget and you have two widgets to the same address both will chance the wrcode to the check mark, this is why the callback is called twice.

lissavxo (commented):
Ready to review

react/lib/util/api-client.ts
```diff
@@ -10,7 +10,31 @@ import {
...
+export const shouldTriggerOnSuccess = (
+  transaction: Transaction,
+  thisPaymentId?: string,
+  cryptoAmount?: string,
+  thisOpReturn?: string,
+) => {
+  const { amount, paymentId, message } = transaction;
+
+  const formattedTxAmount = resolveNumber(amount);
+  const formattedAmountSet = cryptoAmount ? resolveNumber(cryptoAmount) : false;
```
chedieck, line 24: Maybe call this `formattedExpectedAmount`, or something like that. This function in general is being quite hard to read — the bottom line here  is that there are values that come in the tx and there are values that had been defined in the paybutton, and we want to trigger onSuccess if they match.

A variable of type `BigNumber | boolean` named with a suffix ("Set") that only refers to one of the possible types (the boolean one), conjugated in the past tense, which is a homograph of the infinitive (also "set"),  without expliciting that this refers to the expected value — certainly doesn't help, especially when this suffix is [the word with most meanings in the English language](https://www.guinnessworldrecords.com/world-records/english-word-with-the-most-meanings/).

react/lib/components/Widget/WidgetContainer.tsx
```diff
@@ -132,32 +133,29 @@ export const WidgetContainer: React.FunctionComponent<WidgetContainerProps> =
...
-            }Received ${receivedAmount} ${currencyTicker}`,
-            snackbarOptions,
-          );
-        const txPaymentId = transactionPaymentId
-        const isCryptoAmountValid = (cryptoAmount && receivedAmount.isEqualTo(resolveNumber(cryptoAmount))) || !cryptoAmount;
-        const isPaymentIdValid = thisPaymentId ? txPaymentId === thisPaymentId : true;
-
-        if (isCryptoAmountValid && isPaymentIdValid)
-        {
+        const { amount } = transaction;
```
chedieck, line 136: I don't this this deconstruction adds clarity. This will only be used once below, and in a context where there are two amounts: the expected (named `cryptoAmount`) and the received `transaction.amount`. Not adding this line and using `transaction.amount` below makes more clear that the value being used there is the one received in the transaction.

react/lib/util/api-client.ts
```diff
@@ -10,7 +10,31 @@ import {
...
+  const formattedTxOpReturn = JSON.stringify(message);
+  const formattedOpReturn = JSON.stringify(thisOpReturn);
+
+  const isAmountValid = formattedAmountSet
```
chedieck, line 28: So if `formattedAmountSet` is false or 0 (since it can be a boolean or a BigNumber) the amount is valid? Maybe call this `expectedAndReceivedAmountsMatch` or simply `amountsMatch` or anything that makes it more clear what is being checked here.

chedieck (changes requested):
Added some comments to improve readability.

react/lib/util/api-client.ts
lissavxo: thanks for the feedback, I just gave it a little cleanup to improve reading and also make it less verbose

react/lib/util/api-client.ts
lissavxo: if formattedAmountSet (now -> expectedAmount) is not set it means any amount will be valid and make amount validation true so amount valid.

lissavxo:
Ready to review

chedieck (changes requested):
Tested on my server chedieck.com/paybutton

By trying to pay the first widget visible on the page, both get triggered to success, even though their OPRETURN and amount differs.

This doesn't happen when I use the bundle from master.

lissavxo:
> Tested on my server chedieck.com/paybutton
> 
> By trying to pay the first widget visible on the page, both get triggered to success, even though their OPRETURN and amount differs.
> 
> This doesn't happen when I use the bundle from master.

There was an issue with crypto amount, it was not reaching the widget component, it is fixed, but the function still has a wrong scenario when comparing the OP_RETURN value. 

when op return is set like described here -> 
op-return="classOf=2013 bullYears=2013|2017|2021"
we are receiving the transaction by socket event  like this -> 
{
    "classOf": "2013",
    "bullYears": [
        "2013",
        "2017",
        "2021"
    ]
} 

the api converts the string "classOf=2013 bullYears=2013|2017|2021" for a the object  above. 
I am trying to figure out how to make this validation of opReturn in this case.

react/lib/tests/util/api-client.test.ts
```diff
@@ -0,0 +1,148 @@
+import { shouldTriggerOnSuccess } from '../../util/api-client';
+import { Transaction, resolveNumber } from '../../util';
+
+jest.mock('axios');
+
+global.fetch = jest.fn();
+
+describe('API Client Util Tests', () => {
+  describe('shouldTriggerOnSuccess', () => {
+    it('returns true when all conditions match', () => {
```
Klakurka, line 10: These test names should make sense on their own. This one is super generic and doesn't really tell me anything.

Klakurka (changes requested):

chedieck:
> Tested on my server chedieck.com/paybutton
> 
> By trying to pay the first widget visible on the page, both get triggered to success, even though their OPRETURN and amount differs.

Now what I get is that success is not triggered on neither.

lissavxo:
> > Tested on my server chedieck.com/paybutton
> > By trying to pay the first widget visible on the page, both get triggered to success, even though their OPRETURN and amount differs.
> 
> Now what I get is that success is not triggered on neither.

This was due to rawMessage not being returned from server, when I tested on local server on master it works because local server returned rawMessage. Even thou I did a change to consider the message if there is no rawMessage in transaction. 
So the logic is , if there is a rawMessage we compare with the rawMessage if not we compare with message if exists.

Klakurka:
Conflicts.

react/lib/util/api-client.ts
chedieck, line 1: why should this function go here? it has more to do with the socket.ts file than to this one

chedieck (changes requested):

chedieck:
Same issue as before when I tested. This was tested in my localhost, having paybutton-server running locally.

image

Receiving a 20XEC tx activates success in a paybutton with 20USD of amount

lissavxo:
> Collaborator

Can you provide more info on this ? like the code you used to test it 

> Same issue as before when I tested. This was tested in my localhost, having paybutton-server running locally.
> 
> image
> 
> Receiving a 20XEC tx activates success in a paybutton with 20USD of amount. Even if there is only the 20USD widget there.
> 
> This is wrong not only because the amount is not being taken into account to activate it, but neither is the paymentId.



> Same issue as before when I tested. This was tested in my localhost, having paybutton-server running locally.
> 
> image
> 
> Receiving a 20XEC tx activates success in a paybutton with 20USD of amount. Even if there is only the 20USD widget there.
> 
> This is wrong not only because the amount is not being taken into account to activate it, but neither is the paymentId.

provide the code you used to declare the widgets please

Klakurka (approved):

chedieck (approved):
LGTM!

Klakurka (approved):

PR #405, opened 2024-05-10
refactor: add index utils

lissavxo:
Related to #403 

Description
---
- added index to util folder 
- added small format fixes 

Test plan
---
`yarn build && yarn dev`
also check `yarn test`


react/lib/util/constants.ts
```diff
@@ -0,0 +1,9 @@
+export const FIAT_CURRENCIES = ['USD', 'CAD', 'EUR', 'GBP', 'AUD'];
+
+export const CRYPTO_CURRENCIES = ['BCH', 'XEC'];
+
+export const DECIMALS: { [key: string]: number } = {
+    BCH: 8,
+    XEC: 8,
```
Klakurka, line 7: Why isn't XEC = 2 here?

react/lib/util/constants.ts
```diff
@@ -0,0 +1,9 @@
+export const FIAT_CURRENCIES = ['USD', 'CAD', 'EUR', 'GBP', 'AUD'];
```
Klakurka, line 1: I'm pretty sure we don't support EUR, GBP and AUD yet right?

react/lib/tests/util/currency.test.ts
```diff
@@ -0,0 +1,60 @@
...
+
+  describe('isValidCurrency', () => {
+    beforeEach(() => {
+      jest.resetAllMocks();
+    });
+
+    it('identifies fiat currencies correctly', () => {
+      expect(isValidCurrency('USD')).toBeTruthy();
+      expect(isValidCurrency('EUR')).toBeTruthy();
+    });
```
Klakurka, line 38: Should be false right?

Klakurka (changes requested):
Very big branch - could have broken this up but I'll leave it to @chedieck to decide if it's ok or not this time.

react/lib/util/constants.ts
lissavxo: nice catch, just fixed it and added tests to that file to make sure we don't miss it next time

react/lib/util/constants.ts
lissavxo: @chedieck can confirm, he added this extra currencies  in the previous branch here : commit [8aa0c3fc7c5930ae1b76e8f95df4403b72b8729c](https://github.com/PayButton/paybutton/pull/403/commits/8aa0c3fc7c5930ae1b76e8f95df4403b72b8729c). I did a cherry pick in this commit and added to this branch

lissavxo: this is just replicating the logic that we have on master now,
this constant was just like this on master: 
 
`export const fiatCurrencies = ['USD', 'CAD', 'EUR', 'GBP', 'AUD'] as const;` 

and it is used to verify if it is fiat or not with the function isFiat  and also used to check if the currency is supported in the isValidCurrency function,  so we need to fix the logic

Klakurka: Ok well then let's create a new MR branched off this one to reduce the fait currencies down to just CAD and USD as that's all we currently support.

react/lib/tests/util/currency.test.ts
```diff
@@ -0,0 +1,63 @@
+import { isCrypto, isValidCurrency } from '../../util';
+
+describe('API Client Util Tests', () => {
+  describe('isCrypto', () => {
+    it('recognizes valid crypto currencies', () => {
+      expect(isCrypto('BCH')).toBeTruthy();
+      expect(isCrypto('XEC')).toBeTruthy();
+    });
```
Klakurka, line 8: What happens when we enter 'bch' or 'xec' (lowercase)?

chedieck (approved):

Klakurka (approved):

PR #407, opened 2024-05-17
refactor: add constant util

lissavxo:
Related to #405 

Description
---
Added constants to a separated file and fixed tests

Test plan
---
yarn build && yarn dev

### DIFF
--- a/react/lib/util/api-client.ts
+++ b/react/lib/util/api-client.ts
@@ -3,6 +3,7 @@ import _ from 'lodash';
 import { Socket } from 'socket.io-client'
 import config from '../config.json'
 import { isValidCashAddress, isValidXecAddress } from './address';
+import { CRYPTO_CURRENCIES, FIAT_CURRENCIES } from './constants';
 
 export const getAddressDetails = async (
   address: string,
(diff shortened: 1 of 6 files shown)


react/lib/util/api-client.ts
```diff
@@ -107,25 +108,23 @@ export default {
   getAddressBalance,
 };
 
-export const fiatCurrencies = ['USD', 'CAD'] as const;
 
-type fiatCurrenciesTuple = typeof fiatCurrencies; // readonly ['USD', 'CAD' ]
-export type fiatCurrency = fiatCurrenciesTuple[number]; // "USD" | "CAD" 
+type fiatCurrenciesTuple = typeof FIAT_CURRENCIES;
+export type fiatCurrency = fiatCurrenciesTuple[number];
```
chedieck, line 113: since we're at it, what about renaming this to UpperCamelCase, just as you renamed the constant to UPPER_SNAKE_CASE, following the conventions of not only constants but also types?

react/lib/util/api-client.ts
```diff
@@ -107,25 +108,23 @@ export default {
...
 
-export const cryptoCurrencies = ['BCH', 'XEC'] as const;
-type cryptoCurrenciesTuple = typeof cryptoCurrencies; // readonly ['BCH', 'XEC']
-export type cryptoCurrency = cryptoCurrenciesTuple[number]; // "BCH" | "XEC"
+type cryptoCurrenciesTuple = typeof CRYPTO_CURRENCIES;
+export type cryptoCurrency = cryptoCurrenciesTuple[number];
```
chedieck, line 116: same here

chedieck (changes requested):

chedieck (approved):
Nvm the other comments, I checked the other PR where the types are being reallocated and will suggest that there.

Klakurka (changes requested):
How about adding some appropriate tests?

lissavxo:
> How about adding some appropriate tests?

Is It  necessary to test constants file?  it is just defining configuration, no logic implemented, when we test other utils these lines will be covered by jest 
Besides adding the constants file we changed imports in the components that used it, I will see how I can write some test to cover that changes

Klakurka (approved):
OK

PR #409, opened 2024-05-17
refactor: add types util

lissavxo:
Related to #405 

Description
---

This adds a file with the types being used in the codebase 

Test plan
---
yarn build && yarn dev

### DIFF
--- a/react/lib/util/types.ts
+++ b/react/lib/util/types.ts
@@ -1,0 +1,146 @@
+import { fiatCurrencies, cryptoCurrencies } from "./api-client";
+
+export type currencyObject = {
+    float: number;
+    string: string;
+    currency: string;
+};
+
+export interface Transaction {
+    hash: string
+    amount: string
+    paymentId: string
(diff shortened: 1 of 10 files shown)


react/lib/util/types.ts
chedieck, line 1: Can we put everything here UpperCamelCase? I think it's an easy change, besides `currencyObject` which will be tricky because, across the codebase, sometimes is a variable and sometime is a type, but that is why we shouldn't even have it camelCased in the first place.

chedieck (changes requested):
Left a comment, also:

In order to avoid having to deal with conflicts later on, I'd suggest that you  e.g merge #407 into this branch and put in the description that it depends on #407 being merged so it can be reviewed.

If you don't want to do that it's fine of course, but after merging #407 and merging master into here, there are gonna be conflicts that will need to be solved, so in a way this depends in #407 being solved anyway.

Klakurka (approved):

PR #411, opened 2024-05-22
Chore/add tests coverage

lissavxo:
Related to #405 

Description
---
Set up tests coverage command

Test plan
---
`yarn build && yarn dev` 
to test coverage run: 
`yarn jest:coverage`
and check folder `react/coverage`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ node_modules
 dist
 .env.local
 paybutton/dev/demo/paybutton.js
+react/coverage
--- a/package.json
+++ b/package.json
@@ -9,6 +9,7 @@
   "scripts": {
     "dev": "cd paybutton && yarn dev",
     "test": "cd react && yarn test",
+    "test:coverage": "cd react && yarn test:coverage",
     "build:react": "cd react && yarn && yarn build",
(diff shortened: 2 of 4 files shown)


Klakurka:
Typo in instructions; should be: `yarn test:coverage`

Klakurka (changes requested):
I get some errors running `yarn test:coverage`

image

Klakurka:
Is it possible to also have it load the index.html of the test coverage in the default browser when you run the command? If that's too annoying for others then I'd be ok if it was a different command but for me it would be useful rather than having to manually go view it after.

I know that it shows a lot of detail in the command line output but the HTML one is better.

lissavxo:
> I get some errors running `yarn test:coverage`
> 
> image

we had set 80% of coverage Threshold, we do not have 80% of code covered by tests at this point so it throws this error, I just removed the config so we don't have this error

Klakurka (approved):
I think having the min threshold for coverage is kinda useless... ideally we just keep working until it's 100% and then maintain 100% forever...

chedieck (approved):

PR #412, opened 2024-05-22
Chore/add socket util

lissavxo:
Related to #405 

Description
---
Moved socket logic from the API client to a new utility file called `socket.ts`. 
The reason for moving this function is to isolate Its logic. The logic of `setListener` is outside the scope of the API client's responsibilities. In terms of testing, when we write tests for the API client, we would have to check both the API logic (such as Axios calls and error handling) and the socket logic (such as if it correctly creates the socket listener and handles incoming transactions) This would result in a larger test file with tests of different scopes.

Test plan
---
`yarn build && yarn dev `
also check
`yarn test`


Klakurka:
It's really helpful to put the "why" of the branch in the description too.

Klakurka (approved):

PR #414, opened 2024-06-04
feat: add build:dev script

lissavxo:
Related to #403 

Description
---
Added a script to run `yarn build` command in react project  and also `yarn dev` to restart dev environment.
This command helps development because it rebuilds react so the changes made in code can be seen in the development page with just one command

Test plan
---
run `yarn build:dev`


lissavxo:
> I don't understand the naming very well. `yarn dev` will run the bundle in refresh mode, but `dev:refresh` will first build react and then run `yarn dev`
> 
> What about calling it `dev:rebuild`?



> I don't understand the naming very well. `yarn dev` will run the bundle in refresh mode, but `dev:refresh` will first build react and then run `yarn dev`
> 
> What about calling it `dev:rebuild`?

When we run command  `yarn dev` it builds `paybutton` project and serves it in a development environment at `localhost:10001`. It is called "dev" because it serves a development environment. Command `dev:refresh` would rebuild the `react` package, rebuild the `paybutton` project again and serve the development environment, _refreshing_ it in the process.

I named it this way with the environment in mind. It not only builds the "dev" environment but also serves it

chedieck (approved):

Klakurka (approved):

PR #415, opened 2024-06-04
feat: paybutton generator dev

lissavxo:
Related to #403 

Description
---
This is a generator that starts with a paybutton declared via html and using `Vue js` it gets prop values from a form  and creates a new button for testing 

This generator was created to help test quickly different props in an isolated button

While working on it we found this issues: 

- props are not working properly when creating a paybutton via html (div with paybutton class), for example: 
    - `disable = false ` does disables the button
    - `disabe-payment-id = false ` does disable paymentId

- need to check if all props are taking effect on PayButton.render , for example: 
    - theme prop does not apply the theme 

Test plan
---
Run `yarn dev` when it runs go to localhost:10001 and go to Generator page. 

test the button that is there
try to add some props and generate a new button

### DIFF
--- a/paybutton/dev/demo/index.html
+++ b/paybutton/dev/demo/index.html
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html [object Object]>
+<html>
 
 <head>
   <meta charset="utf-8" />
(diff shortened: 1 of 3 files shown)


chedieck (approved):
Looks good.

Klakurka (changes requested):
- It seems to have trouble generating a 2nd time (clicking the generate button once, then again after changing more stuff).
- 'Editable' is always on even if you turn it off and try generating. I think others have the same issue but it's hard to tell with the 1st issue there.

chedieck:
> * It seems to have trouble generating a 2nd time (clicking the generate button once, then again after changing more stuff).

Hmm, I didn't got that. Can you tell exactly how to reproduce this?
>     * 'Editable' is always on even if you turn it off and try generating. I think others have the same issue but it's hard to tell with the 1st issue there.

I think this has to do with a parallel issue: like how lissa said
> disable = false does disables the button

... the same is happening to "editable". It's off by default, but explicitly setting  `editable=false`  actually makes it editable.

Klakurka:
eg. 
image

Klakurka:
RE issue 1: try updating the button w/ disabled=true then try messing with it after - it's always disabled.

Klakurka (approved):
We'll deal with the issues I ran into in a different branch as they're part of a larger issue.

Issue #187, opened 2022-11-30
Integrate alt currency payments


Klakurka:
We will need to use a service that supports XEC of course.

Klakurka:
Could possibly have a "Don't have any XEC?" button somewhere.

Klakurka:
We can try a different service if SideShift doesn't come though.

Klakurka:
https://documenter.getpostman.com/view/8180765/SVfTPnM8?version=latest#intro

Klakurka:
Obviously we'll want to have the API key coming in from the config file.

Klakurka:
XEC support was just added to sideshift.

Klakurka:
Our affiliate code: `https://sideshift.ai/a/bvc`

Klakurka:
We also need to add some logic so that it will only appear when the XEC amount requested is >= 1,000,000 XEC (1 BCHA).

Klakurka:
https://docs.sideshift.ai/

Klakurka:
https://bch.games/ has a good implementation that would be similar to ours aside from that we will want to do a fixed amount sometimes.

Klakurka:
1. 
![image](https://github.com/PayButton/paybutton/assets/4713204/6ce5320c-5420-48e5-adf2-45621644e35c)
2. 
![image](https://github.com/PayButton/paybutton/assets/4713204/ced3ff7e-412b-4ecf-b7a5-19f2328e63b2)
3. 
![image](https://github.com/PayButton/paybutton/assets/4713204/bab65420-0668-4d26-885c-6b4f4ba768c2)

Klakurka:
Just a reminder too that sideshift says they don't and won't support OP_RETURN so we will need to figure out a solution for this. Since we're using their API, we can get a confirmation of the tx having been sent (by sideshift) which we can use to do the success trigger.

Klakurka:
Also I was thinking that we would introduce a new config option to "prioritize" a particular currency in the UX so by default it would be the same as the currency actually being received by the business.

If they decided to change it to say ETH or w/e, in the UX we would basically make it as "ETH focused" as possible, even though the receiving business might be taking XEC or w/e.

Klakurka:
We can allow people to visit the following link to get support:

`https://sideshift.ai/orders/<SHIFT_ID>?openSupport=true`

Perhaps we will want to save the latest shift IDs in localstorage.

[...]

PR #418, opened 2024-07-01
feat: sideshift integration

chedieck:
Related to #187

Description
---
Adds logic to select coins using SideShift's API, displays their amount and the rate of conversion. 

This PR adds the logic for selecting the amount & currency to pay with, but **it doesn't yet implement the payment itself.** 

E.g, by selecting BTC and clicking "Pay with BTC", nothing will happen aside from a console.log of "WIP".

It's also important to notice that I didn't work in the design of the components, only their functionality. This PR lays ground for a frontend work to be done, with the goal of making the new  stuff more intuitive and agreeable for the eyes. 

Finally, it is also important to notice that this PR is off of a proxy branch, meaning approving and merging it won't commit anything to master (since it's incomplete). Nonetheless, it should be reviewed as if the changes here proposed are good enough (in what they propose to address) to eventually go to master (when all the changes are complete).

Test plan
---
For editable and non-editable paybuttons, there should be a "Don't have any XEC?" text which upon clicking will allow the user to select an alternative cryptocurrency to pay with.

### DIFF
--- a/react/lib/components/Widget/Widget.tsx
+++ b/react/lib/components/Widget/Widget.tsx
@@ -6,6 +6,8 @@ import {
   makeStyles,
   TextField,
   Grid,
+  Select,
+  MenuItem
 } from '@material-ui/core';
 import React, { useEffect, useMemo, useState } from 'react';
 import copy from 'copy-to-clipboard';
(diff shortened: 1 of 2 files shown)


Klakurka (changes requested):
Can't seem to get this to build.

image

Klakurka (changes requested):
- Don't allow people to select the same currency that they're paying to (eg. XEC).

chedieck:
> * Don't allow people to select the same currency that they're paying to (eg. XEC).

addressed.

Klakurka (approved):

Klakurka:
This is one of those situations where it would have been good to put this behind a "feature flag" but I'm also ok just merging it as-is. If we had a bit more going on where we'd want to deploy some other feature/fix before this went live, we'd definitely want that flag there to be able to switch it on once it's actually ready.

Issue #417, opened 2024-06-26
Click-to-copy URI doesn't have amount (at least on BCH)


Klakurka:
You can see the issue live on https://coin.dance/support.

PR #421, opened 2024-08-15
fix: amount missing bch qrcode

lissavxo:
Related to #417 

Description
---
Added logic to add amount to copy url

Test plan
---
Run the project with yarn:clean:build
Create a bch button or widget and copy the qrcode url check if amount is there

### DIFF
--- a/react/lib/components/Widget/Widget.tsx
+++ b/react/lib/components/Widget/Widget.tsx
@@ -7,7 +7,7 @@ import {
   TextField,
   Grid,
 } from '@material-ui/core';
-import React, { useEffect, useMemo, useState } from 'react';
+import React, { useCallback, useEffect, useMemo, useState } from 'react';
 import copy from 'copy-to-clipboard';
 import QRCode, { BaseQRCodeProps } from 'qrcode.react';
 import io, { Socket } from 'socket.io-client';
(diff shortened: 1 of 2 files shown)


react/lib/components/Widget/Widget.tsx
```diff
@@ -535,7 +535,9 @@ export const Widget: React.FunctionComponent<WidgetProps> = props => {
       thisAmount = thisCurrencyObject ? thisCurrencyObject.float : undefined;
     }
     if (isValidCashAddress(address)) {
```
Klakurka, line 537: "isValidCashAddress" is ambiguous... does this function check if it's a valid BCH address?

I think I'd prefer if we had a `isValidAddress` which accepted `address` and `currency` OR replace it with a function that takes in an address and returns the currency it matches (or false if no match). What happens to this section of code when we add a 3rd or 4th currency down the line?

Klakurka (changes requested):

react/lib/components/Widget/Widget.tsx
lissavxo: I believe it is named like this because of this : 
return xecaddr.isCashAddress(address);

I agree with you, will work on refactoring this part

Klakurka (dismissed):

react/lib/components/Widget/Widget.tsx
```diff
@@ -526,31 +504,29 @@ export const Widget: React.FunctionComponent<WidgetProps> = props => {
...
-    } else {
-      thisAmount = thisCurrencyObject ? thisCurrencyObject.float : undefined;
-    }
-    if (isValidCashAddress(address)) {
-      prefixedAddress = `bitcoincash:${address.replace(/^.*:/, '')}`;
-    } else {
-      prefixedAddress = `ecash:${address.replace(/^.*:/, '')}${
-        thisAmount ? `?amount=${thisAmount}` : ''
-      }`;
+  const resolveUrl = (currency: string, amount?: number): string | undefined => {
```
chedieck, line 512: You should use [useCallback](https://react.dev/reference/react/useCallback#usecallback) here: notice how this function use reactive values like `disabled`, `to`, and `opReturn`.

I actually didn't find issues when I tested, but I think they could appear if we don't address that

chedieck (changes requested):

chedieck (changes requested):
When I copy the QR code by clicking on it, the amount is not present, only the op_return_raw.

lissavxo:
> When I copy the QR code by clicking on it, the amount is not present, only the op_return_raw.

Could you try with this config: 
  "wsBaseUrl": "https://socket.paybutton.org",
  "apiBaseUrl": "https://api.paybutton.org",
  
I just gave it a test and it seems to be working

chedieck (approved):
LGTM

PR #423, opened 2024-08-19
fix: editable prop

lissavxo:
Related to #420 

Description
---
Fixed editable prop not working when "false" was given 

Test plan
---
Run the project with yarn clean:build && yarn dev 
create a paybutton with `editable="false"` it should not enable edit box


Klakurka (approved):
Editable is good but Disabled still won't clear properly in the Generator Test page.

If you want to address that in another branch, that's fine with me.

chedieck (approved):

PR #424, opened 2024-08-22
Fix/disabled prop

lissavxo:
Related to #420 

Depends on
---
- [x] #423 

Description
---
Fix widget and button being disabled when the prop is "false"  

Test plan
---
Run yarn `clean:build && yarn dev` , go to the generator page and test the disabled prop

### DIFF
--- a/paybutton/dev/demo/paybutton-generator.html
+++ b/paybutton/dev/demo/paybutton-generator.html
@@ -5,10 +5,11 @@
     <title>Paybutton Generator Test</title>
     <link rel="stylesheet" type="text/css" href="style.css">
     <style>
-       body {
+      body {
         padding: auto;
         font-family: Arial, sans-serif;
       }
+
       input,
       button {
         padding: 8px;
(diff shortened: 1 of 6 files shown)


Klakurka (changes requested):
1. Go to http://localhost:10001/paybutton-generator.html
2. Set disabled = true
3. Click 'Generate'
4. Test button (it'll be disabled)
5. Set disabled = false
6. Click 'Generate'
7. See that it's still disabled (it should be enabled)

lissavxo:
> 1. Go to http://localhost:10001/paybutton-generator.html
> 2. Set disabled = true
> 3. Click 'Generate'
> 4. Test button (it'll be disabled)
> 5. Set disabled = false
> 6. Click 'Generate'
> 7. See that it's still disabled (it should be enabled)

Fixed

Klakurka (changes requested):
The generator is just very broken for me with this build.

Not much seems to apply properly when clicking 'Generate'. `Disabled` definitely doesn't work at all.

Master works fine though so there's something going on with this one.

Klakurka (approved):
Seems to work now but I noticed that the `theme` doesn't apply in the generator.

Klakurka:
Conflicts.

chedieck (approved):
LGTM

PR #431, opened 2024-10-22
[UI] Style edits to sideshift UI

johnkuney:
Related to #

Description
---
Some general UI updates to the sideshift integration 

Test plan
---


Klakurka (changes requested):
We can go further with this IMO:

1. Payment page: 
image. I think we're using fixed rate we can remove "at least" to make it clear for them to just send that exact amount. Let's make the address copyable - same for the SideShift payment ID.

2. If there's only 1 network, which is the case most of the time, let's just hide the network dropdown.

3. Let's change "Trade" to "Send" 
image

4. I've always thought this looks kinda bad so maybe there's something better we can do to improve the value input here.

5. I don't know why there's a page that asks you to enter an amount again after you've already set the XEC amount before clicking "Don't have any XEC?"... should be removed right?

johnkuney:
Okay added most of those. What were you referring to in point 4? The amount input? 

As far as why there is another amount screen, That should only be visible when the altpaymentEditable attribute is set to true, otherwise it should just show the amount

Klakurka:
Yeah sorry - I was talking about the amount input.

Klakurka (changes requested):
Some currencies don't seem to work (no buttons to continue):

image

johnkuney:
hmm yeah not sure whats happening there, I can look at the amount input

Klakurka (commented):
It's a bit better and making it look nicer is out of scope but something we'll want to think about updating later.

image

We can't currently progress throug the wizard because the API is broken but that's not our fault. Sideshift will need to fix that. Once they have it working again, we can re-test this.

Klakurka (approved):
I'm going to merge this for now and we can touch it up after SideShift gets their stuff figured out.

Issue #422, opened 2024-08-16
Have SideShift be opt-in


Klakurka:
We may change this in the future but for now, we'll make it opt-in (as a true/false option).

PR #432, opened 2024-11-06
feat:enable altpayment prop

lissavxo:
Related to #422 

Description
---
Created a new prop to enable altpayment logic 

Test plan
---
Create a widget or a paybutton and test prop enable-altpayment="true" and enable-altpayment="false", this prop should control the sideshift logic, it should work on paybutton and widget


Klakurka (changes requested):
Doesn't build for me:

image

Klakurka (changes requested):
Can you add an entry into the docs? All 3 langs.

Klakurka (changes requested):
On the Chinese docs, the desc format is different from what we have on the other properties:

image
image

Can you use the same tags & quotes as we're using on the other ones to keep them consistent?

Klakurka (approved):

Klakurka (changes requested):
Actually can you add this into `paybutton-generator.html` ?

chedieck (changes requested):
Adding `enable-altpayment="false"` doesn't seem to work at the buttons / widgets of paybutton/dev/demo/index.html

Klakurka (approved):

chedieck (approved):
After running `yarn clean:build`, it worked out well

PR #434, opened 2024-11-13
chore: realocate, rename, uncommit config

chedieck:
Description
---
Realocates the config file to the root dir, so it's not hidden in a place that has to be memorized (it was in `react/lib` before).
Also changes its name from `config.json` to `paybutton-config.json` and leaves it gitignored. An additional file called `example-config.json` was added too, also to the root directory.

Test plan
---
Just run `yarn dev`, the example config should be copied to the new expected location `./paybutton-config.json` and then the software should run as always.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ dist
 .env.local
 paybutton/dev/demo/paybutton.js
 react/coverage
+paybutton-config.json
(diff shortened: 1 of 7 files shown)


Klakurka (changes requested):
Seems to work but there's a test failure.

Klakurka (approved):

Issue #426, opened 2024-09-23
Improve behavior when byte size for user data exceeded


Klakurka:
- When regenerating the widget, have the error text clear (revert back to what it should be showing). 
![image](https://github.com/user-attachments/assets/aa70b224-6c0b-413e-96e3-00e97f9d5f74)
- Blur QR code (like we do when disabled=true).

Issue #327, opened 2024-01-13
Improve error displaying


chedieck:
Right now is not even red, also maybe we should blur the QR code or something?

![image](https://github.com/PayButton/paybutton/assets/21281174/8eca90cf-8ce7-4f28-84bf-83320e65dda0)

Klakurka:
Yeah the QR code should be blurred out -- in fact we already do this in at least one case... I'm just blanking on when that is but we can reuse that for whenever there's an error.

johnkuney:
Is there more to do here? Looks like there is a red error message and the QR is blurred now on an error

PR #435, opened 2024-11-13
Fix/op return fail

lissavxo:
Related to #426 

Description
---
Fix opReturn fail behavior 

Test plan
---
create a button/widget with `op-return` prop set to a large string like this:
`"ABCDEFGHIJKLMNOPABCDEFlkfdnlkfndlksnkjsnfkldsnlknlksssssssss;sssssssssGHIJKLMNOPABCDEFGHIJKLMNOPABCDEFGHIJKLMNOPABCDEFGHIJKLMNOPABCDEFGHIJKLMNOPABCDEFGHIJKLMNOPABCDEFGHIJKLMNOPABCDEFGHIJKLMNOPABCDEFGHIJKLMNOP"`
the expected behavior is blur QR code and show the error message in console instead of in the button text.

### DIFF
--- a/react/lib/components/Widget/Widget.tsx
+++ b/react/lib/components/Widget/Widget.tsx
@@ -458,10 +458,6 @@ export const Widget: React.FunctionComponent<WidgetProps> = props => {
       console.error(err);
       setErrorMsg((err as Error).message);
       setDisabled(true)
-      setTimeout((): void => {
-        setErrorMsg('')
-        setText(`Send any amount of ${addressType}`);
-      }, 5000);
     }
   }, [props.opReturn, paymentId, disablePaymentId]);
 


Klakurka (changes requested):
- Rebase to master.
- Can we add this (OP_RETURN msg) to the button generator?

chedieck (changes requested):
The behavior described in the test plan checks out, however, the issue mentions reverting the text back, not getting ridding of it and using the console.

I don't think showing the error in the console only is a good idea, I think we should display that same text of before but having it cleared out after 5s (as we do in other places).

The color could be red to indicate an error, too.

chedieck (approved):

Klakurka (changes requested):
After a few seconds, the text at the top changes from the red error to "send any amount":

image

I get that this is kinda out of scope and related to #327 but IMO we should just deal with this now. Up to you if you want to do it in this branch or a new one.

chedieck (approved):
> After a few seconds, the text at the top changes from the red error to "send any amount":

I thought this was the desired behavior, as per the issue description.

In any case, I also think is better to leave the error there.

Klakurka:
> > After a few seconds, the text at the top changes from the red error to "send any amount":
> 
> I thought this was the desired behavior, as per the issue description.
> 
> In any case, I also think is better to leave the error there.

It's referring to when the widget is regenerated.

Klakurka (dismissed):
We can improve the error message at some point as the wording is pretty technical but it's fine for now.

Klakurka (approved):

Klakurka (approved):

chedieck (approved):

PR #436, opened 2024-11-18
Feat/generator render widgets

lissavxo:
Related to #426 

Description
---
Created a new field in the `generator.html` to select whether it will render a `paybutton` or a` paybutton-widget`. 

This change will help testing, before generator only rendered buttons now it can be used to widgets too.

Test plan
---
Run `yarn dev` access `http://localhost:10001/paybutton-generator.html` test generating widgets/buttons.

### DIFF
--- a/paybutton/dev/demo/paybutton-generator.html
+++ b/paybutton/dev/demo/paybutton-generator.html
@@ -11,7 +11,8 @@
       }
 
       input,
-      button {
+      button,
+      select {
         padding: 8px;
         border-radius: 6px !important;
         border: 1px solid #ccc;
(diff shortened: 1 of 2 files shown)


Klakurka (approved):

chedieck:
One thing to notice: this branch has commits from #435, but not all of them (is out-of-date)

When including changes from another branch in the PR branch, try to use the `Depends on` in the PR description template. This way is easier for reviewers to know which changes to ignore, or which branch to diff it against, or even for you to remember to merge that branch back into this in case that changes (as happened now)

PR #439, opened 2024-12-07
fix: sideshift button not rendering

chedieck:
Description
---
Fixes issue where an editable amount paybutton wouldn't allow for the user to finish the process of paying with another currency.

Test plan
---
After selecting an altcoin, the button "Send with XYZ" wouldn't appear in master. In this branch, it should appear normally.

Remarks
---
- The change that makes it work is the empty `<div></div>` . I have no idea why is that so. Tried to figure out but decided not to spend too much time on that. My guess is something related to the new CSS, since this worked fine before.
- It seems that the "Don't have any XEC?" text doesn't appear if a button is of fixed amount (i.e; not editable). I will check this out in a next PR.

### DIFF
--- a/react/lib/components/Widget/AltpaymentWidget.tsx
+++ b/react/lib/components/Widget/AltpaymentWidget.tsx
@@ -186,7 +186,7 @@ export const AltpaymentWidget: React.FunctionComponent<AltpaymentProps> = props 
           copiedMessage.style.padding = "5px 0 5px 5px";
           copiedMessage.style.zIndex = "10";
           copiedMessage.style.display = "none";
-  
+
     if (contentElement) {
       const content = contentElement.textContent || "";
       navigator.clipboard.writeText(content);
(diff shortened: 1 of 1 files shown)


react/lib/components/Widget/AltpaymentWidget.tsx
```diff
@@ -414,7 +415,7 @@ export const AltpaymentWidget: React.FunctionComponent<AltpaymentProps> = props
                       <Select
                         labelId="select-coin-label"
                         className={classes.select_box}
-                        value={selectedCoin?.coin}
+                        value={selectedCoin?.coin ?? null}
```
chedieck, line 418: this and the similar line below were put here to fixes console errors that would appear in the browser

Klakurka (changes requested):
Let's pop a code comment on that empty div as that's really weird...

Klakurka:
@johnkuney any idea what's going on there?

chedieck:
> Let's pop a code comment on that empty div as that's really weird...

I had tried to do:
`<div></div> {/* This is necessary for the button rendering */}`

That made the sideshift logo and the input field HUGE. Some weird stuff happening indeed.

chedieck:
I found the change that broke it and reverted to its previous state.

johnkuney:
Cool, any idea why that broke it? just curious, why does it need to be a grid? and why would that effect a button outside the grid?

chedieck:
> Cool, any idea why that broke it? just curious, why does it need to be a grid? and why would that effect a button outside the grid?

I don't know it, really. Maybe it has to do with CSS in the widget, since the AltpaymentWidget is a children. Maybe it has to do with the way react renders stuff and hydrates the DOM in the client side. But truly, those are just guesses from my side.

Klakurka (approved):

Issue #428, opened 2024-09-23
Replace text = empty string with nbsp; to prevent button height from collapsing


Klakurka:
We want to avoid this: 
![image](https://github.com/user-attachments/assets/b30428ed-054c-4307-988d-6045edee3e33)

PR #441, opened 2024-12-12
fix: text empty nbsp

lissavxo:
Related to #428 

Description
---
Added logic to render `<div>&nbsp;</div>` when text is empty string 

Test plan
---
Run the project with yarn dev and test the button with prop text=""


Klakurka (changes requested):
Still getting this on the button generator dev page 
image

Klakurka (approved):

chedieck (approved):

Issue #444, opened 2025-01-11
Add 'tx inputs' to callback arguments


Klakurka:
Right now we have hash, amount, and paymentId but it would be useful to know which address(es) was used for the inputs. Cashtab will (as of this writing) only show 1 but electrum would be >=1.

Klakurka:
We should add this to the payment trigger options on the server as well.

Klakurka:
Server task: https://github.com/PayButton/paybutton-server/issues/916

Klakurka:
I bumped the prior up on this one because we're trying to get an MVP wordpress plugin working that could really use this.

PR #445, opened 2025-01-12
feat: add inputs to tx type

lissavxo:
Related to #444 

Description
---
Added input addresses to Transaction type 

Test plan
---
- Run server in this PR https://github.com/PayButton/paybutton-server/pull/917 
- Run client and set a callback (onSuccess or onTransaction) 
- Check if there are inputAddresses returned in the tx object. 
- Check if the inputAddresses matches the ones in the block explorer

### DIFF
--- a/react/lib/util/types.ts
+++ b/react/lib/util/types.ts
@@ -15,6 +15,7 @@ export interface Transaction {
     timestamp: number
     address: string
     rawMessage?: string
+    inputAddresses?: (string | undefined)[]
 }
 
 export interface UtxoDetails {


Klakurka (dismissed):

react/lib/util/types.ts
```diff
@@ -15,6 +15,7 @@ export interface Transaction {
     timestamp: number
     address: string
     rawMessage?: string
+    inputAddresses?: (string | undefined)[]
```
chedieck, line 18: Same thing I commented in the server PR: I don't see the sense into having `undefined` as a possibility here.

chedieck (changes requested):

chedieck (approved):

Klakurka (approved):

Issue #442, opened 2024-12-18
Improve SideShift UI


Klakurka:
- Make sure it keeps the same div dimensions as the QR code / grey box. Stop it from jumping (eg. when error message appear/disappear).
- Stop allowing non-numeric digits from being entered into the 'Amount' field: 
![image](https://github.com/user-attachments/assets/f6d09214-5e13-414d-961d-59114dcb41c4) - note the quirk on the error message showing 3 errors at the same time.
- Nicer looking "Shift Ready" page (tbh each one can be improved).
- Better looking display of address, copy buttons, etc.

PR #447, opened 2025-01-14
[UI]: Sideshift ui edits

johnkuney:
Related to #442

Description
---
- Prevent non numerics in the input
- Improved copy buttons and shift value displays 
- style tweaks
- improve outer gray box jumping around. This one was is kind of tricky due to some of the widget's responsive size definitions for the qr inner content, which is not present here. So its better, but not perfect 
Test plan
---
preview and try out a shift

### DIFF
--- a/react/lib/components/Widget/AltpaymentWidget.tsx
+++ b/react/lib/components/Widget/AltpaymentWidget.tsx
@@ -180,11 +180,14 @@ export const AltpaymentWidget: React.FunctionComponent<AltpaymentProps> = props 
     const copiedMessage = document.createElement("div");
           copiedMessage.textContent = "Copied!";
           copiedMessage.style.position = "absolute";
-          copiedMessage.style.width = "100%";
-          copiedMessage.style.top = "-5px";
+          copiedMessage.style.width = "calc(100% - 10px)";
+          copiedMessage.style.height = "calc(100% - 20px)";
+          copiedMessage.style.alignItems = "center";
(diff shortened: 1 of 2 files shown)


Klakurka (changes requested):
- Can we somehow ensure that the widget/dialog height remains the same throughout the wizard? Right now, it jumps on each step.
- Can we add some hover animation on the copy buttons?
- I still kinda feel like this should be more "button-y" 
image

johnkuney:
okay for sure, added those

Klakurka (changes requested):
- The dialog height jumping is better but not perfect. We can improve this in a different PR; not high prio.
- If you go to start a shift w/ an editable amount then back out, the 'Edit amount' field will have a ton of extra decimals in there... XEC only supports 2 so I'm not sure where they're coming from: 
image Notice how the QR code and 'Send' text at the top are fine.

johnkuney:
hmm yes I'm seeing the same, and actually it will let you enter more than two decimals at any point, as well as non numeric characters... 

I can see the value it's using, but not sure the best way to fix this. Either way it should probably be handled in another PR. And I think @lissavxo or @chedieck may have a better idea the best way to handle it

Its using the "thisamount" value which is not run through the getCurrencyObject function which has some formatting rules based on the currency...I imagine the naked state was used for rendering reasons perhaps?...but seems like the handleAmountChange function could be improved to sanitize the value more

Klakurka (approved):

Issue #448, opened 2025-01-17
Sideshift sending back extra decimals into input when backing out


Klakurka:
If you go to start a shift w/ an editable amount then back out, the 'Edit amount' field will have a ton of extra decimals in there... XEC only supports 2 so I'm not sure where they're coming from: 
![image](https://github.com/user-attachments/assets/9fd7a000-7d18-444e-acbb-5039617529c6) Notice how the QR code and 'Send' text at the top are fine.

PR #449, opened 2025-01-22
fix: sideshift amount decimals

lissavxo:
Related to #448 

Description
---
Fixed decimals and settleCoin name 
Fixed decimals of settleCoin 

Test plan
---
- Create a paybutton with both BCH/XEC addresses and check the following behavior -> 
go to sideshift > choose a coin > click `Send with sideshift` > set an amount > go back > go back 

check if the amount is set with correct number of decimals.


Klakurka (approved):

Klakurka (approved):
With that last commit I snuck in, works great in all scenarios I could come up with.

chedieck (approved):

PR #456, opened 2025-01-28
ui: Make SideShift container same size as widget 

johnkuney:
Related to #

Description
---
Moving the sideshift content to be absolutely positioned so its always the same exact size as the widget to prevent any jumping 

Test plan
---
Try some shifts


johnkuney:
Okay sorry I misread your issue, and thought you were talking about the sideshift container changing size. 
In any case think I fixed both things here now actually.
Now the sideshift container should be the same exact size as the widget or dialog, and the "dont have any XEC" button should fade in instead of jumping

Klakurka (approved):
Perfect!

Issue #464, opened 2025-02-01
Cashtab.com not opening on desktop when needed


Klakurka:
It just throws an error.

PR #466, opened 2025-02-02
fix: open cashtab web case

lissavxo:
Related to #464 

Description
---
Open cashtab.com when ElectrumABC and Cashtab Extension is not installed 

Test plan
---
Run `yarn dev`, test send button with ElectrumABC and Cashtab Extension not installed


Klakurka (changes requested):
This kills support for Electrum ABC which we don't want to do.

Are we not able to put a error handler on this and open cashtab.com in that event?

Klakurka (approved):
After talking with a few people, consensus is that this is probably the best thing to do.

Issue #227, opened 2023-08-05
Introduce better animations


Klakurka:
Right now stuff just loads in as it needs to but we can certainly introduce some animations on the QR code and/or payment dialogs.

Example: https://github.com/bitjson/qr-code

Klakurka:
https://www.framer.com/motion/

PR #467, opened 2025-02-06
ui: add animations

johnkuney:
Related to #227

Description
---
Adding some subtle animations to the dialog open, qrcode, and buttons

Test plan
---
preview and try it out. Let me know if you see any issues


Klakurka (changes requested):
Awesome!

- I was wondering if we could use this as an opportunity to improve the QR code clicking. Right now there's no animation but a slight CSS border change that makes the whole dialog shift a few pixels. Anything better we an do there?
- Any thoughts on animations on payment success?
- Really I'd call this a nit but when the QR code takes an extra bit of time to load in (when the spinner is up and it's waiting for a server response), the QR code animation doesn't run because it's past the time when it would have completed normally. Able to make it animate in when the actual QR code is rdy or is that too much work?

Klakurka (approved):
Big improvement!

Issue #270, opened 2023-09-15
Add 'transaction-text' to pair with 'on-transaction'


Klakurka:
Right now we use 'success-text' for both which doesn't allow for much flexibility.

After implementing this:

`success-text` runs when `on-success` fires.

`transaction-text` runs when `on-transaction` fires.

PR #471, opened 2025-03-18
feat: transaction text prop

lissavxo:
Related to #270 

Description
---
Added the `transaction-text` prop. The `transaction-text` will be displayed in the widget when a transaction occurs, but the amount does not match the required amount, preventing a success message. 

Test plan
---
- Add transaction-text prop in a paybutton and widget 
- Test transact-text in paybutton generator

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -560,7 +560,7 @@ successText = "Thanks!"
 
 ## transaction-text
 
-> **The ‘transaction-text’ parameter specifies the text displayed when a new transaction occurs.**
+> **The ‘transaction-text’ parameter specifies the text displayed when a new transaction occurs but the amount or OP_RETURN code are incorrect.**
 
 ?> This parameter is optional. Default value is empty. Possible values are any string.
 
--- a/docs/README.md
+++ b/docs/README.md
@@ -558,6 +558,35 @@ successText = "Thanks!"
 
 <!-- tabs:end -->
 
+## transaction-text
+
+> **The ‘transaction-text’ parameter specifies the text displayed when a new transaction occurs.**
+
+?> This parameter is optional. Default value is empty. Possible values are any string.
+
+**Example:**
+<!-- tabs:start -->
+
+#### ** HTML **
+
+```html
+transaction-text="New transaction!"
+```
+
+#### ** JavaScript **
+
+```javascript
+transactionText: 'New transaction!'
+```
+
+#### ** React **
+
+```react
+transactionText = "New transaction!"
+```
+
+<!-- tabs:end -->
+
 ## on-success
 
 > **The ‘on-success’ parameter specifies the callback function that runs upon successful payment.**
@@ -927,6 +956,8 @@ apiBaseUrl: 'https://paybutton.org'
 ```react
 apiBaseUrl = "https://paybutton.org"
 ```
+
+<!-- tabs:end -->
 
 ## disable-altpayment
 
--- a/docs/zh-cn/README.md
+++ b/docs/zh-cn/README.md
@@ -555,7 +555,7 @@ successText = "Thanks!"
 
 ## transaction-text
 
-> **“transaction-text” 参数指定在新交易发生时显示的文本。**
+> **“transaction-text” 参数指定当发生新交易但金额或 OP_RETURN 代码不正确时显示的文本。**
 
 ?> 此参数为选填，默认值为空。可用值为任何字符串。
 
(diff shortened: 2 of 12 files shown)


react/lib/components/Widget/WidgetContainer.tsx
```diff
@@ -187,6 +190,13 @@ export const WidgetContainer: React.FunctionComponent<WidgetContainerProps> =
             onSuccess?.(transaction);
           } else {
             onTransaction?.(transaction);
+            if (transactionText){
+              setNewTxText(transactionText)
+              setTimeout(() => {
+                setNewTxText('')
+              }, 10000)
+            }
```
Klakurka, line 198: I wasn't really expecting this... I think it's better to stick with the standard toast and perhaps we can make a new task to better differentiate them so people don't mistakenly think they paid the full correct amount when they did not.

Klakurka (changes requested):
- Can you update one of the examples in the demo page to use this? Perhaps on the XEC 'Purchase' button that already has a success-text set.
- Let's go ahead and open a PR on the server to get the button generator updated. We can update the wordpress plugin later.

react/lib/components/Widget/WidgetContainer.tsx
Klakurka: On second thought - @chedieck thoughts on this? Maybe John also has some ideas how best to visually communicate the custom message while also making it clear that less than the full amount was paid.

chedieck: Maybe call it "Unexpected payment" or something like that by default? Probably some other more precise adjective would be better.

As for the PR, this `10000` should be a constant instead of a hardcoded number.

chedieck (changes requested):

Klakurka (changes requested):
I tried to send the wrong amount to a widget and wasn't able to get the tx to appear. I even entered the op_return_raw for it but still nothing.

react/lib/components/Widget/WidgetContainer.tsx
```diff
@@ -174,19 +185,27 @@ export const WidgetContainer: React.FunctionComponent<WidgetContainerProps> =
...
               );
 
             setSuccess(true);
             onSuccess?.(transaction);
           } else {
             onTransaction?.(transaction);
+            if (transactionText){
+              enqueueSnackbar(
+                `${
+                  transactionText ? transactionText : 'New transaction'
```
Klakurka, line 203: I don't think this should be set by default.

Let's keep the old behaviour (of no toast) if it's not set.

docs/README.md
```diff
@@ -557,6 +557,35 @@ successText = "Thanks!"
 
 <!-- tabs:end -->
 
+## transaction-text
+
+> **The ‘transaction-text’ parameter specifies the text displayed when a new transaction occurs.**
```
Klakurka, line 562: Let's clarify that it's displayed when a new transaction is received that doesn't have the correct amount & OP_RETURN code.

On BCH we just check that it's the correct amount as OP_RETURN is not really supported well on the wallet side.

docs/zh-cn/README.md
```diff
@@ -553,6 +553,37 @@ successText = "Thanks!"
 
 <!-- tabs:end -->
 
+## transaction-text
+
+> **“transaction-text” 参数指定在新交易发生时显示的文本。**
```
Klakurka, line 558: Update as needed.

docs/zh-tw/README.md
```diff
@@ -553,6 +553,36 @@ successText = "Thanks!"
 
 <!-- tabs:end -->
 
+## transaction-text
+
+> **"transaction-text" 參數指定在新交易發生時顯示的文字。**
```
Klakurka, line 558: Update as needed.

Klakurka (changes requested):

Klakurka:
We'll be adding support for BIP 70 in the near future so this feature may end up being obsolete in most cases.

Klakurka:
Conflicts.

lissavxo:
> Conflicts.

resolved

chedieck (approved):
Working well

Klakurka (changes requested):
I actually couldn't get this to work.

My steps:

1. Update `dev/demo/index.html`, adding `transaction-text="partial payment"` to the `Basic` button.
2. Open http://localhost:10001/
3. Click the `Basic` button to open the QR code.
4. Click the QR code to copy the BIP21 URI. Paste the URI somewhere and edit out the amount. Copy the URI without the amount.
5. Open https://cashtab.com/ and paste the URI into the address field.
6. Enter an amount of 8 (or something that isn't 10).
7. Click `Send`.

When I follow these steps, nothing happens.

Klakurka (approved):

Issue #366, opened 2024-01-29
Ensure that random-satoshis option works even if payment ID isn't present


Klakurka:
This is mainly for the benefit of BCH which doesn't have industry-wide support for op_return.

We want on-success to fire if random-satoshis is active and the amount received is exact.

Klakurka:
Related to https://github.com/PayButton/paybutton-server/issues/819

Klakurka:
Might be fixed? Need to test BCH.

Klakurka:
(tested by Lissa)

PR #482, opened 2025-04-10
fix: randomSatoshis trigger success

lissavxo:
Related to #366 

Description
---
Fix button not triggering success when random satoshis is active 

Test plan
---
1 - Test component(paybutton and widget) with no amount should trigger succes
2 - Test if component (paybutton and widget) with an amount and without random satoshis active is triggering sucess
3 - Test if component (paybutton and widget) with an amount and random satoshis active is triggering success as expected

### DIFF
--- a/react/lib/components/Widget/WidgetContainer.tsx
+++ b/react/lib/components/Widget/WidgetContainer.tsx
@@ -108,8 +108,8 @@ export const WidgetContainer: React.FunctionComponent<WidgetContainerProps> =
       disableSound,
       ...widgetProps
     } = props;
-    const [internalCurrencyObj, _setInternalCurrencyObj] = useState<CurrencyObject>();
-    const setCurrencyObj = props.setCurrencyObj || _setInternalCurrencyObj;
+    const [internalCurrencyObj, setInternalCurrencyObj] = useState<CurrencyObject>();
+    const setCurrencyObj = props.setCurrencyObj || setInternalCurrencyObj;
     const currencyObj = props.currencyObj || internalCurrencyObj;
 
     const [thisPaymentId, setThisPaymentId] = useState<string | undefined>();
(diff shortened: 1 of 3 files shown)


react/lib/components/Widget/WidgetContainer.tsx
```diff
@@ -108,6 +105,9 @@ export const WidgetContainer: React.FunctionComponent<WidgetContainerProps> =
       contributionOffset,
       ...widgetProps
     } = props;
+    const [internalCurrencyObj, _setInternalCurrencyObj] = useState<CurrencyObject>();
```
chedieck, line 108: Setter shouldn't have been prefixed with a underscore, it is used in the line below

chedieck (changes requested):
Not sure if I understood it correctly, but I was guessing `random-satoshis=true` should make ONLY the amounts being equal necessary to call a transaction a success transaction.

i.e. ignoring the opReturn, and with him the paymentId.

chedieck (changes requested):
~I'm not getting a success with the same amount if I don't send anything in the OP_RETURN.~

~Tried for the button:~

```
    <div class="paybutton" to="ecash:qq3zv9s3jna6x5t0fxzfgd2zwteqyyjrscf9zmthz7" text="Mine" amount="21"
                                                                                             random-satoshis="true"
      currency="XEC" goal-amount="1000" success-text="WOW!" on-close="myClose" on-open="myOpen"
      on-success="mySuccessFunction" on-transaction="myTransactionFunction" 
                ></div>
```

react/lib/util/validate.ts
```diff
@@ -8,9 +8,10 @@ export const shouldTriggerOnSuccess = async (
     transaction: Transaction,
     currency: string,
     price: number,
+    randomSatoshis: number | boolean,
```
chedieck, line 11: why 'number'? AFAIK this is a boolean.

chedieck (changes requested):
My bad, didn't pulled the latest changes, it is actually working well. Just left a code question.

react/lib/util/validate.ts
lissavxo: Check WidgetContainerProps randomSatoshis is boolean | number there

chedieck (approved):

Issue #379, opened 2024-02-02
Have BCH ignore OP_RETURN for the purposes of triggering onSuccess


Klakurka:
BCH doesn't have the same level of standardization of wallets as XEC does so we should simply ignore OP_RETURN there.

PR #484, opened 2025-04-15
feat: ignore opReturn bch

lissavxo:
Related to #379 

Description
---
Ignore Op return trigger success validation on bch transactions 

Test plan
---
Check if button is triggering success on BCH transactions with wrong OP_RETURN

### DIFF
--- a/react/lib/util/validate.ts
+++ b/react/lib/util/validate.ts
@@ -1,5 +1,5 @@
 import BigNumber from "bignumber.js";
-import { getCurrencyTypeFromAddress } from "./address";
+import { getAddressPrefix, getCurrencyTypeFromAddress } from "./address";
 import { resolveNumber } from "./number";
 import { Currency, CurrencyObject, Transaction } from "./types";
 import { DECIMALS } from "./constants";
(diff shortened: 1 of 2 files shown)


react/lib/util/validate.ts
```diff
@@ -51,6 +53,15 @@ export const shouldTriggerOnSuccess = (
       const opReturn = rawOpReturnIsEmptyOrUndefined ? message : rawOpReturn
       const opReturnIsEmptyOrUndefined = opReturn === '' || opReturn === undefined;
 
+    if(!isBCH){
```
chedieck, line 56: This is really weird. 
1. Identation is missing
2. The first three lines of the if clause are identical to what has just been done before entering the `if (!isBCH) {...}`
3. The last two lines of the if clause are identical to what will be done after leaving the `if (!isBCH) {...}`

chedieck (changes requested):

react/lib/util/validate.ts
lissavxo: should be good now

Issue #300, opened 2024-01-01
Continue checking for payment after closing dialog


Klakurka:
If a payment is made (same address, amount, OP_RETURN data), still fire the success toast as well as onSuccess/onTransaction just like if they had the dialog open or were using the widget.

The goal here is to increase payment detection reliability from the client's perspective. Some websites may entirely rely on the client to let them know about payments (eg. people who have opted not to create a paybutton.org account) so we want to give them as much of a chance to have those functions fire regardless of what the paying user is doing.

PR #485, opened 2025-04-15
[#300] feat: keep socket open on closed button

chedieck:
Related to #300 

Description
---
Changes the logic in order to make the `onSuccess` function still fire for a closed PayButton, as long as it has been opened once since page load.

Test plan
---
Make sure not only #300 is working well, but also sideshift altpayments & Widgets, since the changes also affect them.


lissavxo:
I tested these cases -> 
1 Button any amount 
    trigger success when opened [yes]
    trigger success when closed [no]
2 Button with amount set 
    trigger success when opened [yes]
    trigger success when closed [no]
3 Widget any amount 
    trigger success [yes]
4 Widget with amount set 
    trigger success [yes]
   
I also noticed altpayment is not working properly It freezes on loading step  
Screenshot 2025-04-16 at 14 35 01

lissavxo:
It is not working properly, I was able to trigger the payment success behavior once but the other times I tried I can only hear the sound the dialog doesn't open

lissavxo (dismissed):

Klakurka (changes requested):
We need to make a decision around how to handle the user clicking multiple buttons on the page before paying.

If I open one button dialog, close it, then open a 2nd different one and pay to that, the first one will open up on top of my active button dialog.

One option is to only ever keep the _last_ connection open. If somebody opens a new dialog, kill the old one.

Another option is to continue having it work as-is but if I pay to button #2, don't trigger #1 to open in my face.

Either option is good with me.

Klakurka (approved):

Klakurka:
Conflicts.

Issue #465, opened 2025-02-02
Add 'successSound' config option


Klakurka:
The default points to the built-in sound.

I'm thinking we can allow them to set any valid path to a sound file.

If they set it to something like a blank string, false, or null then it will effectively be like muting it.

Obviously they could also add this code into the success callback so I'm also open to just doing a 'disable-sound' config option instead -  but we should discuss it before developing.

lissavxo:
I think we could support both `success-sound` and `disable-sound` props. If the provided sound link is invalid, we could fall back to the default sound or mute it. If disable-sound is true, we mute it entirely.

PR #487, opened 2025-04-17
Feat/disable sound prop

lissavxo:
Related to #465 

Description
---
Added a new prop to disable on success sound

Test plan
---
Test if the sound is disabled when using the new prop on success scenario

### DIFF
--- a/react/lib/components/Widget/WidgetContainer.tsx
+++ b/react/lib/components/Widget/WidgetContainer.tsx
@@ -16,7 +16,8 @@ import {
   isValidCurrency,
   resolveNumber,
   shouldTriggerOnSuccess,
-  getCurrencyObject
+  getCurrencyObject,
+  isPropsTrue
 } from '../../util';
 
 import Widget, { WidgetProps } from './Widget';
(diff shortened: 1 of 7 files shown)


react/lib/components/Widget/WidgetContainer.tsx
```diff
@@ -148,7 +151,7 @@ export const WidgetContainer: React.FunctionComponent<WidgetContainerProps> =
         if (altpaymentShift) {
           const shiftStatus = await paymentClient.getPaymentStatus(altpaymentShift.id)
           if (shiftStatus.status === 'settled') {
-            if (sound) txSound.play().catch(() => {});
+            if (sound && disableSound !== true) txSound.play().catch(() => {});
```
chedieck, line 154: shouldn't this be `!isPropsTrue(disableSound)` too?

chedieck (changes requested):

react/lib/components/Widget/WidgetContainer.tsx
lissavxo: fixed

chedieck (approved):
Works well.

Issue #486, opened 2025-04-16
Editable + random satoshis shows two different randomized values


chedieck:
![Image](https://github.com/user-attachments/assets/5ebf47ea-f2e3-4c5a-a9ad-a8007e92d62d)

PR #488, opened 2025-04-17
fix/amount-editable-random-satoshis

lissavxo:
Related to #486 

Description
---
Fix amount on editable field 

Test plan
---
Check if amount showing up on editable input field is the same as the amount in the text, check with and without randomSatoshis active


chedieck (approved):
Great fix.

Issue #230, opened 2023-08-11
Add button size as a config option


Klakurka:
One option is to make specific sizes:

- Small
- Regular (default)
- Large
- X-Large

Another option is to allow entering a number where 100 is the default. Setting it to 10 would mean it's 10% the size.

The parameter name can just be 'size' .

Vidhanvyrs:
Can you please be specific about which buttons are we talking here and where should the changes nedded to be done 
is the changes should be in this 
var config = {
  to: 'bitcoincash:qrmm7edwuj4jf7tnvygjyztyy0a0qxvl7q9ayphulp',
  amount: 4.5,
  goalAmount: 300000,
  canEdit: false,
  currency: 'AUD',
  text: 'Tip Us a Coffee!',
size: regular, // Added this inorder to change it according the class that we choose but from where do they choose?
  hoverText: 'One BCHinno',
  theme: {
    palette: {
      primary: '#ee8b2b',
      secondary: '#fefbf8',
      tertiary: '#504030'
    }
  },
  Like i think the way we can fix this is by getting the button id and making it chages it's class by using switch case we can do that but first we do need to write the css code for classes Small large Regular and xlarge
  but how to apply this change to the button like suppose our id of the button is dotlasher
   var buttonElement = document.getElementById('dotlasher');
  switch (config.size) {
      case 'Small':
        buttonElement.classList.add('small');
        break;
      case 'Large':
        buttonElement.classList.add('large');
        break;
      case 'X-Large':
        buttonElement.classList.add('x-large');
        break;
      default:
        break;
    }

Vidhanvyrs:
Now the code for the css is here but where to add them that is what i am not getting 
 <style>
    .button {
      padding: 10px 20px;
      font-size: 16px;
    }
    
    .button.small {
      padding: 5px 10px;
      font-size: 12px;
    }
    
    .button.large {
      padding: 15px 30px;
      font-size: 20px;
    }
    
    .button.x-large {
      padding: 20px 40px;
      font-size: 24px;
    }
  </style>

PR #489, opened 2025-04-21
Feat/size prop

lissavxo:
Related to #230 

Description
---
Added size prop, the size config will apply a scale to the button making it smaller or larger. Available options are 'xs', 'sm', 'md'(default), 'lg', 'xl'. 

Test plan
---
Test size prop options  'xs', 'sm', 'md', 'lg', 'xl'

### DIFF
--- a/react/lib/util/constants.ts
+++ b/react/lib/util/constants.ts
@@ -12,3 +12,5 @@ export const CURRENCY_PREFIXES_MAP: Record<typeof CRYPTO_CURRENCIES[number], str
     bch: 'bitcoincash',
     xec: 'ecash',
 };
+
+export type buttonSize = 'xs' | 'sm' | 'md' | 'lg' | 'xl' | undefined;
--- a/react/lib/util/constants.ts
+++ b/react/lib/util/constants.ts
@@ -13,4 +13,4 @@ export const CURRENCY_PREFIXES_MAP: Record<typeof CRYPTO_CURRENCIES[number], str
     xec: 'ecash',
 };
 
-export type buttonSize = 'xs' | 'sm' | 'md' | 'lg' | 'xl' | undefined;
+export type buttonSize = 'xs' | 'sm' | 'md' | 'lg' | 'xl' | "extrasmall" | "small" | "medium" | "large" | "extralarge" | undefined;
--- a/react/lib/util/constants.ts
+++ b/react/lib/util/constants.ts
@@ -13,4 +13,4 @@ export const CURRENCY_PREFIXES_MAP: Record<typeof CRYPTO_CURRENCIES[number], str
     xec: 'ecash',
 };
 
-export type buttonSize = 'xs' | 'sm' | 'md' | 'lg' | 'xl' | "extrasmall" | "small" | "medium" | "large" | "extralarge" | undefined;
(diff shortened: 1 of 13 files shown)


Klakurka (changes requested):
- The actual container div size needs to be adjusted too (both bigger or smaller as necessary): image
- How about adding some full-word aliases? eg. "extrasmall", "small", "medium", "large", "extralarge".

react/lib/util/constants.ts
```diff
@@ -11,4 +11,6 @@ export const DECIMALS: { [key: string]: number } = {
 export const CURRENCY_PREFIXES_MAP: Record<typeof CRYPTO_CURRENCIES[number], string> = {
     bch: 'bitcoincash',
     xec: 'ecash',
-};
\ No newline at end of file
+};
+
+export type buttonSize = 'xs' | 'sm' | 'md' | 'lg' | 'xl' | "extrasmall" | "small" | "medium" | "large" | "extralarge" | undefined;
```
chedieck, line 16: UpperCamelCase here, since it refers to a type

chedieck (changes requested):
Can you elaborate why this props is passed all the way down to `PaybuttonDialog`,  `WidgetContainer` and `Widget`? I ask because the affected button exists before these components, and not inside them.

lissavxo:
> Can you elaborate why this props is passed all the way down to `PaybuttonDialog`, `WidgetContainer` and `Widget`? I ask because the affected button exists before these components, and not inside them.

This way, I can control which button receives the size prop config. Only the PayButton outside the dialog and widget will receive the size prop.

chedieck (dismissed):

Klakurka (changes requested):
- The div size is still the same no matter what button size is chosen (same point I made before).

chedieck (changes requested):
Conflicts

lissavxo:
resolved

Klakurka (changes requested):
Doesn't build for me:

image

lissavxo:
> Doesn't build for me:
> 
> image

Should be okay now

lissavxo:
> would be possible to change the outer container someway other than by using transform? The problem with transform is that it overlaps the outer stuff: image
> 
> (tested using https://jsfiddle.net/ktu0q4o1, if you wanna experiment there)
> 
> Here is the same thing happening in PayButton (I added blue background color to the outer `<div class="paybutton">`, to show how it expands without pushing stuff in the side) image

Now we are just changing the font size instead of transform

johnkuney:
For sure, I may not be understanding the issue exactly, but I think it's more a problem with the grid css in our index.html file.

The grid columns aren't really responsive to the content width, it's instead fitting as many columns it can that are at least 180px on each row. So if the content is larger it overflows and if its its smaller it doesnt adjust. At least I think thats whats happening

If we move to a flexbox layout everything works as expected. Hope you dont mind, I'll push some tweaks I made to make our preview file layout work, I also added some adjustments to the border radius so it scales with the button size as I thought that was a bit better than 10px on every size

Klakurka (approved):

Klakurka (approved):

chedieck (approved):

PR #508, opened 2025-06-17
Feat/autoclose false default

lissavxo:
Related to #

Description
---
- Set auto close to false by default 
- Added auto close prop to generator
- Removed undoing sucess by default

Test plan
---
Create a button, make a payment, check if the dialog keeps open
Also check if sucess checkmark remains there


lissavxo:
> Now it doesn't close by default, but it also doesn't close with `auto-close="true"`

Fixed

Issue #318, opened 2024-01-11
onSuccess does not trigger if amount is in fiat


chedieck:
Probably fiat amount is being checked against arriving amount (in XEC).

Klakurka:
I can't reproduce this.

Klakurka:
Looks fixed to me.

PR #514, opened 2025-07-02
fix:shouldTriggerOnSuccess amount validate

lissavxo:
Related to 

Description
---
Small fix in shouldTriggerOnSuccess function to get the amount from currencyObject.
Sometimes we call `shouldTriggerOnSuccess` function with `currencyObject` but there is no need to convert the amount to value because the `currencyObject.currency` is crypto currency. This PR checks if `currencyObject.currency` is crypto currency and it converts amount to value only if `currencyObject.currency` is fiat. 

Test plan
---


Klakurka:
Is this tied to any task (eg. #318)?

Issue #524, opened 2025-07-29
Introduce TailwindCSS for eventual migration


Klakurka:
We will do a gradual migration over time but we can start by adding it into the project and slowly iterate from there.

PR #525, opened 2025-07-29
Added TailwindCSS library

Klakurka:
Related to #524

Description
---
Introduces TailwindCSS, including 1 migration (Button.tsx) - with the eventual goal of completely migrating away from Material UI.

Test plan
---
Make sure nothing major is broken - pay close attention to the Button.tsx changes as there are definitely some things that need to be fixed up. Have a look at the demo/index.html page to see some styling regressions (eg. when the text has to go to 2 lines due to being too long).

Remarks
---

v3.4 was chosen because v4.x apparently doesn't work as well with microbundle and might be better at least until we've completely migrated away from Material UI.

### DIFF
--- a/react/lib/components/Button/Button.tsx
+++ b/react/lib/components/Button/Button.tsx
@@ -70,9 +70,8 @@ const getButtonClasses = (_theme: Theme, animation: animation, size: ButtonSize)
     'whitespace-nowrap',
     
     // Enhanced border and appearance
-    'border-2',
+    'border',
     'bg-transparent',
-    'backdrop-blur-sm', // Subtle backdrop blur for depth
     
     // Cursor
     'cursor-pointer',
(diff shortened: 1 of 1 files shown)


johnkuney (changes requested):
I tried adding a test div with some tailwind classes and the styles did not come through.
Just added it in the widget component and did yarn build (in react dir) and then yarn dev (in paybutton dir) 

Messed with it for a bit to try to get it to work but was not able to get any tailwind styles to come through

johnkuney:
I was only able to get the tailwind styles (of a test div) to work in the demo if I add an import tailwinds.css in the index.html. Is that the expected behavior? 

I dont quite follow the tailwind.md....or the commands arent working for me. Should npm run build (in react) and then npm run dev (in paybutton) make any tailwinds styles added to components work in the demo? or does it need the npm run watch:tailwind also? I tried them all I believe

Klakurka:
OK - ready again for testing.

`Button.tsx` was mostly migrated over which you can use as an example.

Issue #516, opened 2025-07-06
Switch to cashtab-connect (lib) for interfacing w/ cashtab


Klakurka:
This standardizes how we interact with both the cashtab browser extension and website.

https://www.npmjs.com/package/cashtab-connect

PR #526, opened 2025-08-03
Switched to the new cashtab-connect lib

Klakurka:
Related to #516

Description
---
This change standardizes how we interact with both the Cashtab extension and (fallback) website.

Test plan
---
Initiate payments with the Cashtab extension installed and not installed.

### DIFF
--- a/react/lib/util/cashtab.ts
+++ b/react/lib/util/cashtab.ts
@@ -1,0 +1,95 @@
+import { 
+  CashtabConnect, 
+  CashtabExtensionUnavailableError,
+  CashtabAddressDeniedError,
+  CashtabTimeoutError 
+} from 'cashtab-connect';
+
+// Create a single instance to be reused throughout the app
+const cashtab = new CashtabConnect();
+
+/**
+ * Check if the Cashtab extension is available
+ * @returns Promise<boolean> - true if extension is available, false otherwise
+ */
+export const getCashtabProviderStatus = async (): Promise<boolean> => {
+  try {
+    return await cashtab.isExtensionAvailable();
+  } catch (error) {
+    return false;
+  }
+};
+
+/**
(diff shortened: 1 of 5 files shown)


react/lib/util/cashtab.ts
```diff
@@ -0,0 +1,100 @@
...
+};
+
+/**
+ * Open Cashtab with a BIP21 payment URL
+ * @param bip21Url - The BIP21 formatted payment URL
+ * @param fallbackUrl - Optional fallback URL if extension is not available
+ */
+export const openCashtabPayment = async (bip21Url: string, fallbackUrl?: string): Promise<void> => {
+  try {
+    const isAvailable = await getCashtabProviderStatus();
```
BytesOfMan, line 60: this approach -- checking status on button click -- is probably effective in most current contexts. However it would be wasteful for any application where multiple clicks are expected.

The cashtab provider status is not expected to change during a user's visit (and, even if it did --- e.g. if the user installs the cashtab extension --- a refresh is required for this to be detected).

Would be more robust and payments would be faster if this check were performed on page load. This would also allow you to conditionally render the button, i.e. show the user that it will open Cashtab or the mobile app or Electrum before a click.

Klakurka: Good idea - now we can make this nice little text change on widget load:

image

Klakurka:
Changes:

- Cashtab extension check only happens once (regradless of how many buttons/widgets are on the page) and at page load.
- "Send with XEC wallet" text now shows "Send with Cashtab" if extension is detected.

chedieck (changes requested):
Works well, just the same thing about too many redundant comments pollution

Klakurka (commented):
Some of the comments I do find useful but I removed the rest.

PR #544, opened 2025-08-29
fix: base64 of success audio

chedieck:
Related to #

Description
---

Test plan
---
See if the encoded base64 is indeed the expected file, e.g. by running
`mpv <(cat PATH/TO/success.mp3.json  | jq -r '.base64' | cut -d, -f2 | base64 -d)`


Klakurka:
Could we add a test for this to make sure that the mp3 stays in sync with the json file?

Klakurka (approved):

Issue #574, opened 2025-11-12
Opt-in PB donation w/ payment via heart button


Klakurka:
To not piss anyone off, let's add a small button in the bottom of the dialog/widget somewhere that configures the bonus BIP21 output for the PB donation %.

Possible idea:
![Image](https://github.com/user-attachments/assets/44d4999d-1d21-4ed8-a920-53f1fd937108)

Of course the default will be set by the button config.

Also, let's save this in local storage or something if we can so that if they set it once on a site then it will be on/off by default when they come back or make another payment (remember the % too if on).

PR #575, opened 2025-11-14
feat: Donation opt in/edit button

johnkuney:
Related to #574

Description
---
Adding the user controlled button and input to edit the donation amount 
- heart icon on/off toggle
- editable % amount
- tooltip 
- localstorage persistence 
- only shows when an amount is present and for xec 

Test plan
---
Test a button with an amount and try changing the value, turning it on and off. check the proper donation is being applied 

## Summary by CodeRabbit

* **New Features**
  * Footer controls let users enable/disable donations and set a donation rate (%) via a toggle and numeric input; input auto-enables on positive values and auto-disables on zero.
  * Donation settings persist across sessions and restore previous rate when re-enabled; rates are clamped to valid bounds.
  * Donations apply only when enabled and meet currency/minimum thresholds; displayed amounts and generated links reflect donation adjustments; UI shows tooltips and conditional currency-specific controls.

* **Style**
  * Added fade-slide-up animation and updated footer styling.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/react/lib/components/Widget/Widget.tsx
+++ b/react/lib/components/Widget/Widget.tsx
@@ -249,6 +249,12 @@ export const Widget: React.FunctionComponent<WidgetProps> = props => {
   const price = props.price ?? 0
   const [hasPrice, setHasPrice] = useState(props.price !== undefined && props.price > 0)
   
+  // Helper to clamp donation rate to valid range (1-99 if > 0, or 0)
+  const clampDonationRate = useCallback((value: number): number => {
+    if (value <= 0) return 0
+    return Math.max(1, Math.min(99, value))
+  }, [])
+
   // Load donation rate from localStorage on mount
   const getInitialDonationRate = useCallback(() => {
     if (typeof window !== 'undefined' && window.localStorage) {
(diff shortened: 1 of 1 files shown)


Klakurka (changes requested):
- "Send ___ XEC" text doesn't change when the donation % changes
- Just do whole numbers (no decimals) to make it simpler and take up less horizontal space
- How about putting it on the same line as the "Powered by PayButton.org" so the dialog/widget doesn't need to be any taller?
- Have it off by default if no local storage value is set
- When off, don't display the input box + % label

johnkuney:
Cool added those and made it more subtle

Klakurka (changes requested):
- Can the up/down arrows be shifted over to the right just slightly? There seems to be a bit of room left without having to adjust the width of the whole div. 
image

- Let's limit it to be 0-99 to keep it fitting in the div.

Klakurka (dismissed):

react/lib/components/Widget/Widget.tsx
```diff
@@ -115,6 +117,8 @@ interface StyleProps {
   copied: boolean
 }
 
+const DONATION_RATE_STORAGE_KEY = 'paybutton_donation_rate'
```
chedieck: Move this to react/lib/util/constants.ts, otherwise is still just a hardcoded string in a component file.

Klakurka (changes requested):

Klakurka (dismissed):

